PR #110, opened 2025-05-06
[LLZK-238] Documentation tweaks and automation

iangneal:
- Added some minor doc updates 
- Update CI to test building of docs
- Add workflows to build latest docs for pushes to main (from PRs) and update release workflow to add versioned docs
- Add nix docs target for CI

### DIFF
--- a/.github/workflows/deploy-github-pages.yml
+++ b/.github/workflows/deploy-github-pages.yml
@@ -1,23 +1,34 @@
-name: CI
+name: Deploy GitHub Pages
 on:
-  push:
-    branches: [ main ]
-  pull_request:
+  workflow_call:
+    inputs:
+      version:
+        description: "Version to prepare documentation for. Unset version will update 'latest' version."
+        required: false
+        type: string
+        default: ''
 
 jobs:
-  build:
-    runs-on: ubuntu-24.04
-    timeout-minutes: 10
+  deploy-gh-pages:
+    runs-on: ubuntu-latest
     steps:
-      - name: Checkout
+      - name: Check out repository code
         uses: actions/checkout@v4
         with:
           persist-credentials: false
+
+      - name: Checkout github pages branch
+        uses: actions/checkout@v4
+        with:
+          path: gh-pages
+          ref: gh-pages
+
       # v37.1.0
       - uses: cachix/install-nix-action@d1ca217b388ee87b2507a9a93bf01368bde7cec2
         with:
           install_url: https://releases.nixos.org/nix/nix-2.23.3/install
           github_access_token: "${{ secrets.GITHUB_TOKEN }}"
+
       - name: Set Git credentials
         run: |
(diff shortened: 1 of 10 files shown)


.github/workflows/deploy-github-pages.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Move built documentation to latest
+        if: ${{ inputs.version == '' }}
+        run: mv result/html/* gh-pages
+
+      - name: Move documentation to version folder
+        if: ${{ inputs.version != '' }}
+        run: |
+          mkdir gh-pages/${{ inputs.version }}
+          mv result/html/* gh-pages/${{ inputs.version }}
```
Technius, line 54: The substitutions need to be quoted. I suggest saving the version to an env var like this:
```suggestion
        env:
          VERSION: ${{ inputs.version }}
        run: |
          mkdir gh-pages/"$VERSION"
          mv result/html/* gh-pages/"$VERSION"
```

.github/workflows/deploy-github-pages.yml
```diff
@@ -0,0 +1,55 @@
...
+
```
Technius, line 55: The docs don't seem to actually be uploaded here. I believe you either need to commit the changes back to the `gh-pages` branch (for manual branch-based deployments) or use the [deploy action](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#creating-a-custom-github-actions-workflow-to-publish-your-site).

iangneal: Oops, forgot to add that.

iangneal: Should now build and deploy properly (we'll see)

.github/workflows/deploy-github-pages.yml
```diff
@@ -55,3 +55,24 @@ jobs:
...
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+
+    runs-on: ubuntu-latest
+    steps:
+      - name: Deploy artifact
+        id: deployment
+        uses: actions/deploy-artifact@v4
```
Technius, line 78: Does this handle the case of versioned pages? It seems to me that this will overwrite the entire site on deployment.

iangneal: Tweaked this so that `main` gets its own folder. So now each version will go in its own directory, so adding a new version will add it to the existing set of directories in the `gh-pages` clone, so the artifact will retain all versions. This requires a setting change to serve out of `main` instead of the root, though. I've asked @kferles to do this on #111

iangneal: Alright, actually found a different workaround.

tim-hoffman (approved):
LGTM

PR #112, opened 2025-05-07
Fix formatting issue in code spans

0xddom:
Doxygen, for compatibility reasons, doesn't handle code spans with ticks in them with the normal Markdown syntax. If a tick is present in the code span needs to be surounded with two backticks.


iangneal (approved):
LGTM!

PR #122, opened 2025-05-22
[VAN-269] Fix location sorting bug

iangneal:
- `unknown` locations were not handled properly by the analysis result map, resulting in unintended conflicts that caused results to be dropped (hence the missing analysis results)
- Also adding a small utility method to expose the `mlir::DataFlowSolver` used in a particular analysis so that uses may query lattice values at a given location.


tim-hoffman (approved):
LGTM

PR #126, opened 2025-05-28
add more helpers to produce strings

tim-hoffman:
- use these helpers in all relevant locations
- also refactor debug implementation into its own file

### DIFF
--- a/lib/Util/Debug.cpp
+++ b/lib/Util/Debug.cpp
@@ -1,0 +1,86 @@
+//===-- Debug.cpp -----------------------------------------------*- C++ -*-===//
+//
+// Part of the LLZK Project, under the Apache License v2.0.
+// See LICENSE.txt for license information.
+// Copyright 2025 Veridise Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#include "llzk/Util/Debug.h"
+#include "llzk/Util/StreamHelper.h"
+
+using namespace mlir;
+
+namespace llzk {
+namespace debug {
+
+void dumpSymbolTableWalk(Operation *symbolTableOp) {
+  std::string output; // buffer to avoid multi-threaded mess
+  llvm::raw_string_ostream oss(output);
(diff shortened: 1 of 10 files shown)


lib/Util/Debug.cpp
```diff
@@ -0,0 +1,86 @@
...
+        << op->getAttrOfType<StringAttr>(SymbolTable::getSymbolAttrName()) << '\n';
+  };
+  SymbolTable::walkSymbolTables(symbolTableOp, /*allSymUsesVisible=*/true, walkFn);
+  llvm::outs() << output;
+}
+
+void dumpSymbolTable(llvm::raw_ostream &stream, SymbolTable &symTab, unsigned indent) {
+  for (unsigned i = 0; i < indent; ++i) {
+    stream << "  ";
+  }
```
iangneal, line 33:
```suggestion
  stream.indent(indent);
```
https://llvm.org/doxygen/classllvm_1_1raw__ostream.html#a8fdf5cdf041c8aded7e3308c1c3efacc

lib/Util/Debug.cpp
```diff
@@ -0,0 +1,86 @@
...
+  stream << "Dumping SymbolTable [" << &symTab << "]: \n";
+  auto *rawSymbolTablePtr = reinterpret_cast<char *>(&symTab);
+  auto *privateFieldPtr =
+      reinterpret_cast<llvm::DenseMap<Attribute, Operation *> *>(rawSymbolTablePtr + 8);
+  for (llvm::detail::DenseMapPair<Attribute, Operation *> &p : *privateFieldPtr) {
+    for (unsigned i = 0; i < indent; ++i) {
+      stream << "  ";
+    }
```
iangneal, line 41: Same here, can use the `llvm::raw_ostream::indent` method.

lib/Util/Debug.cpp
tim-hoffman: nice, didn't know that existed

iangneal (approved):
LGTM!

PR #131, opened 2025-06-06
GitHub->Jira Issue syncing

iangneal:
Added workflows for updating Jira with selectively linked issues from GitHub:
- Adding the `sync-to-jira` label will "link" the issue to Jira.
- Adding comments or transitioning the state of a synced issue will make mirrored updates to Jira
- Unsyncing the issue will post a comment to the effect on Jira.

Note that only maintainers can update labels on issues to interact with issue syncing.


.github/workflows/sync-issue-to-jira.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: Setup python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: "pip" # caching pip dependencies
+
+      - name: Install python requirements
+        run: |
+          pip install --upgrade pip && pip install -r scripts/requirements.txt
```
Technius, line 21: This file won't exist because the repository is not checked out.

.github/workflows/sync-issue-to-jira.yml
```diff
@@ -0,0 +1,25 @@
...
+
+      - name: Run link script
+        run: |
+          python3 jira_driver.py --issue ${{github.event.issue.number}} link
```
Technius, line 25: Need to explicitly set the environment variables to their corresponding values in secrets.

.github/workflows/update-synced-issue.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Install python requirements
+        if: steps.check_label.outputs.has-label == 'true'
+        run: |
+          pip install --upgrade pip && pip install -r scripts/requirements.txt
+
+      - name: Run update script
+        if: steps.check_label.outputs.has-label == 'true'
+        run: |
+          python3 jira_driver.py --issue ${{github.event.issue.number}} update --summary "Issue comment ${{github.event.action}}: ${{github.event.comment}}""
```
Technius, line 35: We should assign the github substitutions like `${{github.event.action}}` to environment variables, otherwise we could have injection problems.

Technius (changes requested):
* Permissions should be set to the minimum required: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token
* Is it possible to have one or two workflows instead of four? There's a lot of duplicated logic between them.

iangneal:
Pinging @Technius and or @tim-hoffman to look at this again when there's time.

tim-hoffman:
> Pinging @Technius and or @tim-hoffman to look at this again when there's time.

LGTM but I'm not an expert on github workflows and security practices there so I think it would be best to have final approval from @Technius

tim-hoffman (approved):
LGTM

PR #151, opened 2025-07-03
Prevent hidden warnings

tim-hoffman:
* Correct usage of `SYSTEM` in cmake files so that warnings are not ignored within LLZK header files but are ignored within MLIR and LLVM headers.
* Fix warnings that are now reported within header files.


0xddom:
> There is still one issue we need to address with warnings being reported in generated files from the `r1cs` dialect. There's also the ever-present "#warning _FORTIFY_SOURCE" but I don't know if there's a good way to hide those.

Removing the explicit `-O0` should be enough for the latter

iangneal (approved):
LGTM!

iangneal (approved):
Still LGTM

PR #153, opened 2025-07-24
[LLZK-299] [LLZK-221] Add interval identification for multi-value patterns

iangneal:
Identifies intervals of the form `(s - c_0) * ... * (s - c_N) = 0`, which constrain `s` to be a value within `[c_0, c_N]`. This pattern is commonly used to constrain signals to be binary (e.g., `s * (s - 1) = 0`) or similar. If such a pattern is "sparse", meaning that all constants are not specified in a range (e.g., `s * (s-2) = 0`), the interval will be overapproximated (e.g., `s` in [0, 2] even though the `(s-1)` term was not included in the constraint) since the analysis does not support multiple intervals per value. 

This PR also refactors the interval analysis files, which were becoming too large.


include/llzk/Analysis/Field.h
```diff
@@ -0,0 +1,75 @@
...
+#include <llvm/ADT/APSInt.h>
+#include <llvm/ADT/DenseMap.h>
+#include <llvm/Support/SMTAPI.h>
+
+#include <string_view>
+
+namespace llzk {
+
+/// @brief Information about the prime finite field used for the interval analysis.
+/// @note Seem implementation of initKnownFields for supported primes.
```
tim-hoffman, line 21:
```suggestion
/// @note See implementation of initKnownFields for supported primes.
```

iangneal:
@tim-hoffman the interval documentation is copied verbatim from Picus, so I'll tag @shankarapailoor as well to clarify some things.

PR #156, opened 2025-07-31
[LLZK-301] [LLZK-215] Better `@compute` support for interval analysis

iangneal:
- Add support for more `@compute` operations
- Propagate intervals inferred on `@constrain` inputs to `@compute` function inputs. This allows users to examine how output intervals are computed.
- Refactor `ConstrainRef` generation methods


lib/Analysis/Intervals.cpp
```diff
@@ -407,8 +420,76 @@ llvm::APSInt Interval::width() const {
...
+  );
+  ensure(lhs.isBoolean() && rhs.isBoolean(), "operation only supported for boolean-type intervals");
+  const auto &field = rhs.getField();
+
+  if (lhs.isBoolFalse() && rhs.isBoolFalse()) {
+    return Interval::False(field);
   }
+  if (lhs.isBoolTrue() || rhs.isBoolTrue()) {
+    return Interval::True(field);
+  }
```
tim-hoffman, line 439: This doesn't seem right for AND

lib/Analysis/IntervalAnalysis.cpp
```diff
@@ -526,8 +599,11 @@ ExpressionValue IntervalDataFlowAnalysis::performBinaryArithmetic(
                  .Case<MulFeltOp>([&](MulFeltOp _) { return mul(smtSolver, lhs, rhs); })
                  .Case<DivFeltOp>([&](DivFeltOp divOp) { return div(smtSolver, divOp, lhs, rhs); })
                  .Case<ModFeltOp>([&](ModFeltOp _) { return mod(smtSolver, lhs, rhs); })
-                 .Case<CmpOp>([&](CmpOp cmpOp) {
-    return cmp(smtSolver, cmpOp, lhs, rhs);
+                 .Case<CmpOp>([&](CmpOp cmpOp) { return cmp(smtSolver, cmpOp, lhs, rhs); })
+                 .Case<AndBoolOp>([&](AndBoolOp _) { return boolAnd(smtSolver, lhs, rhs); })
+                 .Case<OrBoolOp>([&](OrBoolOp _) { return boolOr(smtSolver, lhs, rhs); })
+                 .Case<XorBoolOp>([&](XorBoolOp _) {
+    return boolXor(smtSolver, lhs, rhs);
```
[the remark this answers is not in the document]
iangneal: Oh good point.

lib/Analysis/IntervalAnalysis.cpp
```diff
@@ -152,15 +181,15 @@ ExpressionValue neg(llvm::SMTSolverRef solver, const ExpressionValue &val) {
...
-  const auto &f = val.getField();
-  if (val.i.isDegenerate()) {
-    if (val.i == Interval::Degenerate(f, f.zero())) {
-      res.i = Interval::Degenerate(f, f.one());
-    } else {
-      res.i = Interval::Degenerate(f, f.zero());
-    }
-  }
-  res.i = Interval::Boolean(f);
+  // TODO: reason about this slightly better
```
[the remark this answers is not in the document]
iangneal: I addressed it here.

tim-hoffman (dismissed):
LGTM

shankarapailoor:
I don't understand "Propagate intervals inferred on @constrain inputs to @compute function inputs. This allows users to examine how output intervals are computed." Why does propagating intervals inferred on @constrain inputs tell users how output intervals are computed? I'm also not sure if it is sound to do that -- it makes sense if we are assuming all constraints in @constrain are asserts in compute which I'm not sure always holds.

iangneal:
> I don't understand "Propagate intervals inferred on @Constrain inputs to @compute function inputs. This allows users to examine how output intervals are computed." Why does propagating intervals inferred on @Constrain inputs tell users how output intervals are computed?

For example, if `@constrain` constrains input `foo` to be nonzero, then that range constraint is applied to the input to `@compute`.  This is applied because a witness with `foo = 0` would generate an invalid proof.

> I'm also not sure if it is sound to do that -- it makes sense if we are assuming all constraints in @Constrain are asserts in compute which I'm not sure always holds.

Even if the asserts aren't in `@compute`, if the `@compute` function were given inputs outside of that range, it would generate an invalid proof. 

The question I'm trying to answer here is, "given conforming inputs, are computed outputs also conforming?" So this kind of analysis is useful in this case. I'll add a flag to control this behavior, though, since I suppose it not be desirable for all use cases.

test/Analysis/interval_analysis/edwards2montgomery.llzk
```diff
@@ -335,3 +336,18 @@ module attributes {veridise.lang = "llzk"} {
 // CHECK-NEXT:     %arg0[@v][@$super] in Entire
 // CHECK-NEXT:     %arg0[@v][@reciprocal] in Entire
 // CHECK-NEXT: }
+
+// CMP-LABEL: @Div StructIntervals {
+// CMP-NEXT:     compute {
+// CMP-NEXT:         %arg0 in Entire
+// CMP-NEXT:         %arg1 in TypeC:[ 1, 2013265920 ]
+// CMP-NEXT:         %self[@$super] in Entire
```
shankarapailoor, line 344: I would like to see more LLZK tests that exercise different operations in the compute. The PR is quite large so more tests would be great here like bitwise ops, boolean ops, control flow, etc.

lib/Analysis/IntervalAnalysis.cpp
```diff
@@ -521,14 +593,17 @@ ExpressionValue IntervalDataFlowAnalysis::performBinaryArithmetic(
...
+                 .Case<DivFeltOp>([&](auto divOp) { return div(smtSolver, divOp, lhs, rhs); })
+                 .Case<ModFeltOp>([&](auto _) { return mod(smtSolver, lhs, rhs); })
+                 .Case<CmpOp>([&](auto cmpOp) { return cmp(smtSolver, cmpOp, lhs, rhs); })
+                 .Case<AndBoolOp>([&](auto _) { return boolAnd(smtSolver, lhs, rhs); })
+                 .Case<OrBoolOp>([&](auto _) { return boolOr(smtSolver, lhs, rhs); })
+                 .Case<XorBoolOp>([&](auto _) {
+    return boolXor(smtSolver, lhs, rhs);
+  }).Default([&](auto *unsupported) {
     unsupported->emitWarning(
         "unsupported binary arithmetic operation, defaulting to over-approximated intervals"
```
shankarapailoor, line 608: For the compute, I think we should do something with bitwise And since that is super common. Like if you are performing bitwise and with a constant of the form $$2^n-1$$

shankarapailoor (commented):
A few comments and I'd like to see some more tests on the compute side. I also think we should support bitwise and since that is common in a lot of zkvm code.

lib/Analysis/IntervalAnalysis.cpp
```diff
@@ -93,17 +97,52 @@ mod(llvm::SMTSolverRef solver, const ExpressionValue &lhs, const ExpressionValue
...
-  res.i = Interval::Boolean(lhs.getField());
+  const Field &f = lhs.getField();
+  // Default result is any boolean output for when we are unsure about the comparison result.
+  res.i = Interval::Boolean(f);
   switch (op.getPredicate()) {
   case FeltCmpPredicate::EQ:
     res.expr = solver->mkEqual(lhs.expr, rhs.expr);
-    res.i = lhs.i.intersect(rhs.i);
+    if (lhs.i.isDegenerate() && rhs.i.isDegenerate() && lhs.i == rhs.i) {
+      res.i = Interval::True(f);
```
tim-hoffman, line 134: Should both sides degenerate but `!=` give a `False` result here. Similar for the `NE` case below.

tim-hoffman (approved):
LGTM

PR #165, opened 2025-09-16
[LLZK-255] Prohibit modules and structs from being nested within functions

iangneal:
See title.

Pending on #160 to avoid possible merge conflicts.


iangneal:
> We should also check that `StructDefOp` does not contain a `ModuleOp`. 

This is already checked in `StructDefOp,` as only `struct.field` and `function.def` operations are permitted within `struct.def`. I'll push some tests that demonstrate this.

tim-hoffman:
> > We should also check that `StructDefOp` does not contain a `ModuleOp`.
> 
> This is already checked in `StructDefOp,` as only `struct.field` and `function.def` operations are permitted within `struct.def`. I'll push some tests that demonstrate this.

Ah, forgot about that. Thanks!

PR #169, opened 2025-09-19
fix: `unifyDynamicSize` flag was not used to guard the checks

tim-hoffman:


iangneal (approved):
LGTM with the test addition.

PR #172, opened 2025-09-25
minor fix and code quality

tim-hoffman:
Adds the restriction on StructDefOp that its regions cannot have arguments. This just closes a possible source of unexpected stuff happening in the IR since such arguments are not expected to be used in LLZK IR.


shankarapailoor:
I trust that this looks good. Can you add a comment to the PR explaining why the old version was incorrect?

shankarapailoor (approved):

PR #174, opened 2025-09-26
Update from llzk-nix-pkgs and cleanup a bit

tim-hoffman:
- pull in latest llzk-nix-pkgs
  - use modern package names (hyphenation instead of underscores)
  - add build type to package names
  - remove unnecessary `llvm` dev shell


iangneal (approved):
LGTM!

PR #180, opened 2025-09-30
elaborate documentation for ArrayType and StructType parameters

tim-hoffman:
The major point I wanted to make is that `IntegerAttr` parameters have `IndexType` and thus must be representable in the bitwidth of `IndexType`. Additional checks are forthcoming in https://github.com/Veridise/llzk-lib/pull/176.


iangneal (approved):
LGTM

PR #184, opened 2025-10-01
simplify and autoformat the flake definition

tim-hoffman:


iangneal (approved):
LGTM. What did you use to autoformat? Could it be something we add to CI/pre-commit to keep the flake in a nice state?

tim-hoffman:
> LGTM. What did you use to autoformat? Could it be something we add to CI/pre-commit to keep the flake in a nice state?
`nixfmt`
I see some git integrations you could try: https://github.com/NixOS/nixfmt?tab=readme-ov-file#git-hooksnix

PR #186, opened 2025-10-01
run clang-format

tim-hoffman:
A few file seem to have missed the format check after LLVM 20 update


tim-hoffman:
> Do we need to change the formatting checks in the CI to make sure we catch these? Did CI not complain?

I had the same thought so I scanned through the CI check and it looks like it only checks formatting in files edited by the current PR. So what I think happened is these files were never touched during the LLVM 20 upgrade (which upgraded clang-format as well) so they were never run against the format check in CI.

shankarapailoor:
Ok good to know!

shankarapailoor (approved):
LGTM

PR #191, opened 2025-10-15
Code cleanup and documentation clarification in `FuncDefOp::verify()`

tim-hoffman:
- simplify code by using `foreach` loops
- remove the name checks (they are redundant since those are the only names allowed in a struct and we already have the `isInStruct()` check)
- clarify documentation


tim-hoffman:
> Is the AffineMapAttr restriction on all struct functions something that might change at any point? e.g. allowing for private non-`@compute`/`@constrain` methods that are more relaxed?

I don't think there's any reason to ever allow additional functions within structs. The concept of function visibility (public/private) is not present in LLZK.

PR #192, opened 2025-10-15
fix CI build warnings

tim-hoffman:


include/llzk/Analysis/SourceRef.h
```diff
@@ -198,18 +198,18 @@ class SourceRef {
   }
 
   llvm::DynamicAPInt getConstantFeltValue() const {
-    ensure(isConstantFelt(), __FUNCTION__ + mlir::Twine(" requires a constant felt!"));
+    ensure(isConstantFelt(), mlir::Twine(__FUNCTION__).concat(" requires a constant felt!"));
```
0xddom: There's a constructor in `Twine` that apparently is optimized for this particular case:

```c++
mlir::Twine(mlir::StringRef(__FUNCTION__), "....")
```

Constructs the `Twine` in such a way that basically avoids calling `concat`.

Docs: https://llvm.org/doxygen/classllvm_1_1Twine.html#aba6d9549c0ddfa25919915044a81eaf3

flake.nix
```diff
@@ -57,11 +56,15 @@
...
-          };
+          llzk-debug =
+            (final.callPackage ./nix/llzk.nix {
+              clang = final.clang_20;
+              mlir_pkg = final.mlir-debug;
+              cmakeBuildType = "Debug";
+            }).overrideAttrs
+              (attrs: {
+                NIX_CFLAGS_COMPILE = (attrs.NIX_CFLAGS_COMPILE or "") + " -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=0";
+              });
```
0xddom: Another way of getting rid of this warning is to set `-O1` in debug builds or not set it at all. I learned that some time after I wrote those flags in the cmake file way back when. In the zirgen I didn't set the optimization level in `Debug` and the warning went away.

PR #194, opened 2025-10-17
[LLZK-320] Fix docsite TOC, generate better links

iangneal:
- Doxygen upgrade (from LLVM upgrade) broke doxygen-awesome TOC, which has been fixed
- Change autogenerated link style from numeric to github style


tim-hoffman (approved):
LGTM

PR #195, opened 2025-10-20
Cleanup several things in unittests directory

tim-hoffman:
- remove excess brackets/spacing
 - add pretty-printing for DynamicAPInt in test names
 - use cleaner TestWithParam API instead of Test+WithParamInterface
 - remove unused includes
 - refactor to remove redundant tests (there were multiple test suites using the same fixture names which caused an N*M duplication of tests)


iangneal (approved):
LGTM

PR #196, opened 2025-10-22
[LLZK-339] [LLZK-340] Fix analysis caching, interval analysis convergence for yield operations

iangneal:
- Identified and fixed a crash caused by improper caching of analysis results. Analysis results are now keyed to analysis context objects (i.e., to the parameters of the analysis), so multiple invocations of the same analysis will use cached results.
- Fixed an issue with yield operations updating a prior lattice state instead of a succeeding lattice state, causing an update loop that lead to non-convergence of the analysis.
- Added a regression test case.

### DIFF
--- a/include/llzk/Analysis/AnalysisWrappers.h
+++ b/include/llzk/Analysis/AnalysisWrappers.h
@@ -118,7 +118,8 @@ concept ContextType = requires(const Context &a, const Context &b) {
 /// require any context.
 struct NoContext {};
 
-/// @brief Any type that is a subclass of `StructAnalysis`.
+/// @brief Any type that is a subclass of `StructAnalysis` and provided a
+/// `Context` that matches `ContextType`.
 template <typename Analysis, typename Result, typename Context>
 concept StructAnalysisType = requires {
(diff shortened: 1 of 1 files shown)


include/llzk/Analysis/AnalysisWrappers.h
```diff
@@ -118,12 +133,17 @@ concept StructAnalysisType =
...
-  using ResultMap = std::map<
+
+  /// @brief Per-struct results mapping.
+  using StructResults = std::map<
       component::StructDefOp, std::reference_wrapper<const Result>,
       NamedOpLocationLess<component::StructDefOp>>;
 
+  /// @brief A map of this module's structs to the result of the `StructAnalysis` on that struct
+  /// for a given configuration (denoted by the `Context` object).
+  /// The inner `ResultMap` is implemented as an ordered map to control sorting order for iteration.
```
tim-hoffman: Should it be this?
```suggestion
  /// The inner `StructResults` is implemented as an ordered map to control sorting order for iteration.
```

tim-hoffman (approved):
LGTM

PR #197, opened 2025-10-22
removing library linking cycle to allow downstream dynamic linking

tim-hoffman:
Remove cycle in libraries to allow `llzk-rs` to use dynamic linking in dev environment to speed up rebuild time.


iangneal (approved):
Now that we have the ⛓️ 🪵, we're all good to go!

PR #201, opened 2025-11-03
Add isa check for undef.undef

0xddom:
Adds a CAPI function for checking ops against this type.


tim-hoffman:
> LGTM, just needs a changelog

And `cpp` impl of the function 😬

PR #203, opened 2025-11-05
updating nix flakes to import pcl

shankarapailoor:
This is an initial PR which just adds the PCL Dialect as an optional dependency. It updates the nix flakes and llzk.nix to pull the dialect in.


tim-hoffman:
Also, I wonder if the flake import of PCL could also be conditioned on the env var because it still builds PCL regardless.

PR #204, opened 2025-11-05
[LLZK-373] Analysis Improvements

iangneal:
- General: Reduce the number of values stored in dataflow lattices (in both the `SourceRef` and `IntervalAnalysis` lattices) by looking up values in prior lattices rather than propagating all values arbitrarily. This dramatically improves performance for large circuits.
- General: make the `getSovler` helper function return a mutable `DataFlowSolver`, which is required for most operations on the solver due to its internal caching.
- Interval Analysis: update handling of `cast` operations.

### DIFF
--- a/lib/Analysis/ConstraintDependencyGraph.cpp
+++ b/lib/Analysis/ConstraintDependencyGraph.cpp
@@ -139,7 +139,7 @@ mlir::LogicalResult SourceRefAnalysis::visitOperation(
       prior = getLattice(getProgramPointAfter(defOp));
     }
     // Get the value (if there was a defining operation), or the default value.
-    operandVals[operandVal] = prior->getOrDefault(operand.get());
+    operandVals[operandVal] = prior->getOrDefault(operandVal);
   }
 
   // Add operand values, if not already added. Ensures that the default value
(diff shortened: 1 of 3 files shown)


shankarapailoor:
So is the idea with the lattice change that a lattice at a location will only contain entries for values defined at that location? If the interval analysis infers that a value `%foo` has a refined interval then it will update the lattice at %foo's definition?

iangneal:
> So is the idea with the lattice change that a lattice at a location will only contain entries for values defined at that location? If the interval analysis infers that a value `%foo` has a refined interval then it will update the lattice at %foo's definition?

Effectively, yes (the lattices also store their direct operands and the function argument values in the case of the interval analysis). I have a ticket for refactoring these analyses to use the MLIR sparse dataflow analysis (as it is likely ultimately better suited for these use cases), but there will be some more work in doing that (e.g., ensuring that symbol lookups function correctly and handling dependencies for non-result producing operations, which is part of why we initially used the dense analysis), so it is TODO for now. 

Tickets: https://veridise.atlassian.net/browse/LLZK-371, https://veridise.atlassian.net/browse/LLZK-372

lib/Analysis/ConstraintDependencyGraph.cpp
```diff
@@ -395,13 +399,23 @@ mlir::LogicalResult ConstraintDependencyGraph::computeConstraints(
...
       SourceRef prefix(fn.getArgument(i));
-      SourceRefLatticeValue val = lattice->getOrDefault(fnCall.getOperand(i));
+      // Look up the lattice that defines the operand value first, but default
+      // to the afterCallLattice if the operand is not defined by an operand.
+      const SourceRefLattice *operandLattice = afterCallLattice;
+      Value operand = fnCall.getOperand(i);
+      if (Operation *defOp = operand.getDefiningOp()) {
+        ProgramPoint *defPoint = solver.getProgramPointAfter(defOp);
+        operandLattice = solver.lookupState<SourceRefLattice>(defPoint);
+        ensure(afterCallLattice, "could not find lattice for call operand");
```
tim-hoffman:
```suggestion
        ensure(operandLattice, "could not find lattice for call operand");
```

tim-hoffman (approved):
LGTM

PR #208, opened 2025-11-06
Remove unnecessary overlay

iangneal:
The `pcl-mlir-pkg` overlay is already added to the default overlay, so it doesn't need to be applied again.


Technius (approved):
Seems to build, so LGTM

PR #212, opened 2025-11-12
Raghav/inline structs refactor2

raghav198:


include/llzk/Transforms/LLZKInlineStructsPass.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
```
tim-hoffman: missing file header

shankarapailoor (approved):
Approving on behalf of @tim-hoffman

PR #214, opened 2025-11-13
Restrict visibility of some extra class declarations

tim-hoffman:
The functions for custom printing and parsing rules do not need to be public, they are only used within the current op implementation.


iangneal (approved):
LGTM

PR #215, opened 2025-11-14
use unwrapped clang to make include headers accessible

tim-hoffman:
The `llzk-tblgen` tool needs to use unwrapped/dev version of clang to include header files.


iangneal (approved):
LGTM

PR #218, opened 2025-11-14
[NFC] Refactor inline structs pass to prepare for handling more "remaining uses"

tim-hoffman:
_No Functional Changes_

Preparing to fix a bug found in `llzk-inline-structs` pass... Refactoring `handleRemainingUses()` into a class to more clearly split up implementation into functions and avoid passing around the common data. Also added some debug output, clarified/added some documentation, and fixed a couple typos.

Empty changelog because it's only private API and doc changes.

### DIFF
--- a/lib/Transforms/LLZKInlineStructsPass.cpp
+++ b/lib/Transforms/LLZKInlineStructsPass.cpp
@@ -542,9 +542,9 @@ public:
     }
 
     LLVM_DEBUG({
-      llvm::dbgs() << "[handleRemainingUses] op: " << *op << '\n';
-      llvm::dbgs() << "[handleRemainingUses]   in function: " << op->getParentOfType<FuncDefOp>()
-                   << '\n';
+      llvm::dbgs() << "[DanglingUseHandler::handle] op: " << *op << '\n';
+      llvm::dbgs() << "[DanglingUseHandler::handle]   in function: "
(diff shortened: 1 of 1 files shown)


lib/Transforms/LLZKInlineStructsPass.cpp
```diff
@@ -69,29 +69,63 @@ using SrcStructFieldToCloneInDest = std::map<StringRef, DestCloneOfSrcStructFiel
...
+/// (using the given callback) if there is not exactly once such `FieldWriteOp`.
+static FailureOr<FieldWriteOp>
+findOpThatStoresSubcmp(Value writtenValue, function_ref<InFlightDiagnostic()> emitError) {
+  FieldWriteOp foundWrite = nullptr;
+  for (Operation *user : writtenValue.getUsers()) {
+    if (FieldWriteOp writeOp = llvm::dyn_cast<FieldWriteOp>(user)) {
+      // Find the write op that stores the created value
+      if (writeOp.getVal() == writtenValue) {
+        if (foundWrite) {
+          // Note: There is no reason for a subcomponent to be stored to more than one field.
```
iangneal: It seems we don't have this as a semantic check, since code like this is produced by the zirgen frontend and doesn't cause issues in the up-to-date `llzk-opt.` We might want to consider either adding one (if we want this to always be enforced) or adding a transformation pass for it.

PR #231, opened 2025-12-02
Avoid assertion failure from `SetFuncAllowAttrs` when struct is empty

tim-hoffman:


iangneal (changes requested):
I'm not sure that we should allow empty structs, since an empty struct can never be used---a call to `@compute` or `@product` is required to create a struct (since those are the only places where `struct.new` is allowed). Is there a use case for this?

PR #240, opened 2025-12-08
[LLZK-382] Disallow private member access

iangneal:
`struct.readm` only works for private members for the current struct. Otherwise, only private members can be read.


include/llzk/Dialect/Struct/IR/Ops.td
```diff
@@ -326,6 +331,12 @@ def LLZK_MemberWriteOp : MemberRefOpBase<"writem", [WitnessGen]> {
   let summary = "write value to a struct member";
   let description = [{
     This operation writes a value to a named member in a struct/component.
+<<<<<<< HEAD
+
+    A struct can write its own members. However, writing members of other components
+    is not allowed.
+=======
+>>>>>>> origin/feature/renaming
```
tim-hoffman: unresolved merge

tim-hoffman (approved):
LGTM

PR #257, opened 2026-01-09
add CAPI for loopbounds attribute

tim-hoffman:


lib/CAPI/Dialect/LLZK.cpp
```diff
@@ -21,3 +22,11 @@ MLIR_DEFINE_CAPI_DIALECT_REGISTRATION(LLZK, llzk, LLZKDialect)
 MlirAttribute llzkPublicAttrGet(MlirContext ctx) { return wrap(PublicAttr::get(unwrap(ctx))); }
 
 bool llzkAttributeIsAPublicAttr(MlirAttribute attr) { return llvm::isa<PublicAttr>(unwrap(attr)); }
+
+MlirAttribute llzkLoopBoundsAttrGet(MlirContext ctx, int64_t lower, int64_t upper, int64_t step) {
+  return wrap(LoopBoundsAttr::get(unwrap(ctx), toAPInt(lower), toAPInt(upper), toAPInt(step)));
```
shankarapailoor: Question: Is there an assumption that `lower <= upper`?

PR #266, opened 2026-01-13
Walk func to add product_source attr before inlining

raghav198:
After aligning into `@product`, its still useful to know where each op came from when verifying witness/constraint equivalence; this PR adds a `product_source` attribute that's set to either `"compute"` or `"constrain"`, which the LLEQ verifier can read to determine how to appropriately handle it.


lib/Transforms/LLZKComputeConstrainToProductPass.cpp
```diff
@@ -119,6 +119,15 @@ FuncDefOp ComputeConstrainToProductPass::alignFuncs(
 ) {
   OpBuilder funcBuilder(compute);
 
+  // Add compute/constrain attributes
+  compute.walk([&funcBuilder](mlir::Operation *op) {
```
tim-hoffman: nit
```suggestion
  compute.walk([&funcBuilder](Operation *op) {
```

tim-hoffman (approved):
LGTM, just a couple nits since the namespace is already open

iangneal (approved):

Issue #268, opened 2026-01-14
Fix flattening pass to handle affine_map operands


iangneal:
Test case: test/Transforms/Flattening/unresolved_vector.llzk

When a struct accepts an argument with an affine_map, and that affine_map is flattened, the function type no longer matches, causing a type verification error.

iangneal:
I'll close this since affine maps are disallowed from function arguments.

Just a historical note, the original test case referenced can be found in this closed PR: https://github.com/project-llzk/llzk-lib/pull/269

PR #269, opened 2026-01-14
Remove affine map restriction

iangneal:
Removes the restriction on struct types to allow affine map input arguments.

Flattening pass currently does not work with affine_map parameters, see test/Transforms/Flattening/unresolved_vector.llzk. I've added issue #268 to track this.


tim-hoffman:
Thinking through this suggestion more:
```
poly.template @template_f1 {
  poly.param @N
  poly.param @M
  poly.expr @D = @N + @M

  struct.def...
    function.def @compute(%inp: !array.type<@D x !felt.type>) -> !array.type<@D x !felt.type> {
      // etc
    }
    ...
}
```
I believe it does solve the problem while avoiding this affinemap changes in this PR. The `poly.template` proposal does require a templated struct to appear within a `poly.template` op and the additional `poly.expr` op takes the place of an affine map (or could be defined with one to avoid implementing any expression syntax for the RHS).

@iangneal So, how much is this blocking in the circom frontend? Maybe Expression used as array dimension is not that common and it's worth waiting until `poly.template` is available.

tim-hoffman:
> In short, maybe we should just close this, and in the Circom repo, I can implement the simple case (e.g., `var arr[N]` and not `var arr[N+M]`), and we can leave the complex expressions as a TODO pending the expanded template dialect.

Let's do that.

PR #271, opened 2026-01-16
Refactor affine map operands handling

0xddom:
The new POD dialect adds an op that accepts map operands in the same way `array.new` and `function.call` does. To avoid duplication and boilerplate I grouped together the common arguments and created a builder type that helps filling the data.

The changes are **only** for the C API.

### DIFF
--- a/include/llzk-c/Support.h
+++ b/include/llzk-c/Support.h
@@ -123,7 +123,7 @@ typedef struct LlzkAffineMapOperandsBuilder {
 MLIR_CAPI_EXPORTED LlzkAffineMapOperandsBuilder llzkAffineMapOperandsBuilderCreate();
 
 /// Destroys the struct releasing its resources.
-MLIR_CAPI_EXPORTED void llzkAffineMapOperandsBuilderDestroy(LlzkAffineMapOperandsBuilder builder);
+MLIR_CAPI_EXPORTED void llzkAffineMapOperandsBuilderDestroy(LlzkAffineMapOperandsBuilder *builder);
 
 /// Appends the value ranges to the list of map operands.
 MLIR_CAPI_EXPORTED void llzkAffineMapOperandsBuilderAppendOperands(
--- a/lib/CAPI/Support.cpp
+++ b/lib/CAPI/Support.cpp
@@ -120,11 +120,16 @@ LlzkAffineMapOperandsBuilder llzkAffineMapOperandsBuilderCreate() {
   return builder;
 }
 
-void llzkAffineMapOperandsBuilderDestroy(LlzkAffineMapOperandsBuilder builder) {
-  if (builder.mapOperands) {
-    std::free(builder.mapOperands);
+void llzkAffineMapOperandsBuilderDestroy(LlzkAffineMapOperandsBuilder *builder) {
+  if (!builder) {
+    return;
   }
-  maybeDeallocArray(&builder);
+  if (builder->mapOperands) {
+    std::free(builder->mapOperands);
+  }
+  maybeDeallocArray(builder);
+  // Reset values to 0/NULL.
+  *builder = llzkAffineMapOperandsBuilderCreate();
 }
 
 void llzkAffineMapOperandsBuilderAppendOperands(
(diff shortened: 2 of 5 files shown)


lib/CAPI/Support.cpp
```diff
@@ -68,3 +74,104 @@ void mlirOperationWalkReverse(
...
+}
+
+void llzkAffineMapOperandsBuilderAppendOperandsWithDimCount(
+    LlzkAffineMapOperandsBuilder *builder, intptr_t n, MlirValueRange const *mapOperands,
+    int32_t const *dimsPerMap
+) {
+  intptr_t nDimsPerMap = builder->nDimsPerMap < 0
+                             ? unwrap_cast<DenseI32ArrayAttr>(builder->dimsPerMap.attr).size()
+                             : builder->nDimsPerMap;
+  (void)nDimsPerMap; // To silence unsused variable warning if the assert below is compiled out.
```
shankarapailoor:
```suggestion
  (void)nDimsPerMap; // To silence unused variable warning if the assert below is compiled out.
```

include/llzk-c/Support.h
```diff
@@ -96,6 +96,84 @@ void mlirOperationWalkReverse(
...
 );
 
+//===----------------------------------------------------------------------===//
+// Helper types and functions for map operands constructor arguments.
+//===----------------------------------------------------------------------===//
+
+/// Encapsulates the arguments related to affine maps that are common in operation constructors that
+/// support them.
+typedef struct LlzkAffineMapOperandsBuilder {
+  intptr_t nMapOperands;
```
raghav198: Maybe similar to Shankara's comment elsewhere, but is there a reason the # of map operands and dims need to be signed instead of `size_t`?

include/llzk-c/Support.h
```diff
@@ -96,6 +96,84 @@ void mlirOperationWalkReverse(
...
+  MlirValueRange *mapOperands;
+  /// Set to a negative number to indicate that `dimsPerMap.attr` must be used
+  /// instead of `dimsPerMap.array`.
+  /// The default mode is `Array` with an uninitialized array.
+  intptr_t nDimsPerMap;
+  union {
+    MlirAttribute attr;
+    /// List of dimension counts. Is owned by this struct and it gets automatically
+    /// allocated/deallocated if the user switches to or from `Attr` mode.
+    int32_t *array;
```
raghav198: Same question here; does a negative dimension make sense?

lib/CAPI/Support.cpp
```diff
@@ -68,3 +74,104 @@ void mlirOperationWalkReverse(
...
+) {
+
+  auto attr = llzkAffineMapOperandsBuilderGetDimsPerMapAttr(*builder, context);
+  llzkAffineMapOperandsBuilderSetDimsPerMapFromAttr(builder, attr);
+}
+
+MlirAttribute llzkAffineMapOperandsBuilderGetDimsPerMapAttr(
+    LlzkAffineMapOperandsBuilder builder, MlirContext context
+) {
+  if (builder.nDimsPerMap < 0) {
```
raghav198: Ah, if a negative size is being used here as a sentinel value, I think that should be documented up above

changelogs/unreleased/dani__unified-mapops-capi.yaml
```diff
@@ -0,0 +1,13 @@
+# This is a template that lists the available options when creating a changelog entry file
+# Edit according to your changes, before marking a PR as ready for review.
+
```
tim-hoffman: nit
```suggestion
```

include/llzk-c/Support.h
0xddom: The internal representation in the final op of that data uses `DenseI32ArrayAttr` so it's gotta be a `int32_t`

lib/CAPI/Support.cpp
```diff
@@ -68,3 +74,104 @@ void mlirOperationWalkReverse(
...
+  builder.nMapOperands = 0;
+  builder.mapOperands = nullptr;
+  builder.nDimsPerMap = 0;
+  builder.dimsPerMap.array = nullptr;
+
+  return builder;
+}
+
+void llzkAffineMapOperandsBuilderDestroy(LlzkAffineMapOperandsBuilder builder) {
+  if (builder.mapOperands) {
```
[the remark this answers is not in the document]
0xddom: I was thinking of types that are opaque pointers like `MlirOperation`, `MlirContext`, etc. I checked how the other auxiliary structs are handled and they are passed as pointers to the function that consumes/destroys them.

PR #276, opened 2026-01-21
Relax restrictions on bool operators

tim-hoffman:
Removing these restrictions will solve https://github.com/project-llzk/circom/issues/298 and allow the solution suggested in https://github.com/project-llzk/circom/pull/317#discussion_r2709234722


shankarapailoor (approved):
LGTM

PR #277, opened 2026-01-21
265 fuse computeconstrain loops in product program when possible

raghav198:
This contains a pretty basic implementation, but it still needs testcases


include/llzk/Util/Constants.h
```diff
@@ -31,4 +31,7 @@ constexpr char FUNC_NAME_PRODUCT[] = "product";
 /// Name of the attribute on the top-level ModuleOp that specifies the IR language name.
 constexpr char LANG_ATTR_NAME[] = "veridise.lang";
 
+/// Name of the attribute on aligned product program ops that specifies where they came from
```
iangneal:
```suggestion
/// Name of the attribute on aligned product program ops that specifies where they came from.
```

lib/Util/AlignmentHelper.cpp
iangneal: Should remove this empty file

Issue #294, opened 2026-01-29
Using "Main" as the main component name conflicts with Circom conventions


iangneal:
We use "Main" as the internal name for the main component (https://github.com/project-llzk/llzk-lib/blob/5baf809c3130992a9cab182cbbaec2a9007d9541/include/llzk/Util/Constants.h#L23). However, a number of Circom circuits define a `template Main`, which causes a conflict when we try to generate the LLZK main component. We should add a way to differentiate between any user-defined "Main"s and LLZK's special Main component.

One option could be to put LLZK's main in a nested module.

PR #295, opened 2026-01-29
Specify main via attribute instead of reserved name

tim-hoffman:
This changes the means of specifying the main entry point struct from a reserved name "Main" to an attribute on the top module. Also changes the parameter type restriction on the main struct from `!struct.type<@Signal>` to `!felt.type` (and comments that Signal is deprecated, soon to be removed).

Consequently, resolves https://github.com/project-llzk/llzk-lib/issues/294


iangneal (changes requested):
Overall LGTM, but can we also add a test case for using a struct with parameters as the main component?

PR #301, opened 2026-02-04
Add `llzk.fields` attribute

### LINKED ISSUE #135: Support configurable prime field for `felt` data type and conversion
Allow specifying arbitrary primes in a dictionary on the module using an attribute that specifies the defining info for a prime. The felt type can then be parameterized with of these keys. User must implement various conversion functions if they use more than one prime.

Also allow an unspecified-prime felt. It can be decided by the backend. Certain transformations/conversions cannot happen if the prime is unspecified.

Another option is to use a `TypeInterface` for field element types and then any type that implements that interface can be used as a field element.

iangneal:
Resolves #135.

- [x] Update CAPI

### DIFF
--- a/include/llzk-c/Constants.h
+++ b/include/llzk-c/Constants.h
@@ -31,7 +31,7 @@ extern const char *LLZK_LANG_ATTR_NAME;
 
 /// Name of the attribute on the top-level ModuleOp that defines prime fields
 /// used in the circuit.
-extern const char *FIELD_ATTR_NAME;
+extern const char *LLZK_FIELD_ATTR_NAME;
 
 /// Name of the attribute on the top-level ModuleOp that specifies the type of the main struct.
 /// This attribute can appear zero or one times on the top-level ModuleOp and is associated with
--- a/include/llzk-c/Dialect/Felt.h
+++ b/include/llzk-c/Dialect/Felt.h
@@ -28,27 +28,81 @@ extern "C" {
 
 MLIR_DECLARE_CAPI_DIALECT_REGISTRATION(Felt, llzk__felt);
 
-/// Creates a llzk::felt::FeltConstAttr.
+//===----------------------------------------------------------------------===//
+// FeltConstAttr
+//===----------------------------------------------------------------------===//
+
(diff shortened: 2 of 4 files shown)


include/llzk/Dialect/Felt/IR/Types.td
```diff
@@ -20,13 +20,41 @@ include "mlir/Interfaces/MemorySlotInterfaces.td"
 def LLZK_FeltType : TypeDef<FeltDialect, "Felt"> {
   let mnemonic = "type";
   let summary = "finite field element";
-  let description = [{}];
+  let description = [{
+    An element of a finite field. The field element is either of the specified
+    field (e.g., !felt.type<"bn254">) or of an unspecified field (!felt.type)
+    to be filled in by any backend translations.
```
tim-hoffman:
```suggestion
    An element of a finite field. The field can be specified (e.g., !felt.type<"bn254">)
    or left unspecified (!felt.type), to be filled in by the backend consumer of LLZK IR.
    Leaving the field unspecified may reduce the power of certain transformation passes.
```

lib/Dialect/Felt/IR/Dialect.cpp
```diff
@@ -26,11 +28,96 @@
...
+
+  // Custom logic: cache the field, reporting an error if there's a conflict
+  auto errFn = [&odsParser]() {
+    return InFlightDiagnosticWrapper(odsParser.emitError(odsParser.getCurrentLocation()));
+  };
+  Field::addField(fieldNameAttrRes.value(), primeRes.value(), errFn);
+
+  return odsParser.getChecked<FieldSpecAttr>(
+      odsLoc, odsParser.getContext(), mlir::StringAttr((*fieldNameAttrRes)),
+      llvm::APInt((*primeRes))
```
tim-hoffman:
```suggestion
      odsLoc, odsParser.getContext(), mlir::StringAttr(*fieldNameAttrRes), llvm::APInt(*primeRes)
```

tim-hoffman (approved):
LGTM (I pushed one more commit and have one nit here)

PR #404, opened 2026-04-08
v1 to v2 IR migration

tim-hoffman:
- [x] bump version number
- [x] migration for `StructDefOp` (removed parameters, convert to template)
- [x] migration for `FeltConstAttr` (storing FeltType instead of StringAttr with prime field)
- [x] migration for `CallOp` (added optional template parameters)
- [x] add migration test with version 1.x.x bytecode input


include/llzk/Dialect/Shared/TypeConversionPatterns.h
```diff
@@ -0,0 +1,257 @@
...
+
+/// Pattern for `CreateArrayOp`, which lacks the general builder.
+class CreateArrayOpClassReplacePattern : public mlir::OpConversionPattern<array::CreateArrayOp> {
+public:
+  /// Defaults to benefit 0 so that any op-specific pattern at the standard
+  /// benefit (1), or higher, takes priority over this generic fallback.
+  CreateArrayOpClassReplacePattern(mlir::TypeConverter &converter, mlir::MLIRContext *ctx)
+      : mlir::OpConversionPattern<array::CreateArrayOp>(converter, ctx, 0) {}
+
+  mlir::LogicalResult match(array::CreateArrayOp op) const override {
```
iangneal, line 180: Is there a reason this one has a separate `match` and `rewrite` instead of a `matchAndRewrite`?

lib/Dialect/Function/IR/Ops.cpp
```diff
@@ -416,6 +418,83 @@ LogicalResult ReturnOp::verify() {
...
+    return failure();
+  }
+
+  if (failed(reader.readAttribute(prop.mapOpGroupSizes))) {
+    return failure();
+  }
+
+  if (failed(reader.readOptionalAttribute(prop.numDimsPerMap))) {
+    return failure();
+  }
```
iangneal:
```suggestion
  if (failed(reader.readAttribute(prop.callee)) || 
      failed(reader.readAttribute(prop.mapOpGroupSizes)) ||
      failed(reader.readOptionalAttribute(prop.numDimsPerMap))) {
    return failure();
  }
```

lib/Dialect/Function/IR/Ops.cpp
```diff
@@ -416,6 +418,83 @@ LogicalResult ReturnOp::verify() {
...
+    auto readProp = [&]() {
+      if (reader.getBytecodeVersion() >= /*kNativePropertiesODSSegmentSize=*/6) {
+        return reader.readSparseArray(MutableArrayRef(propStorage));
+      };
+      return success();
+    };
+    if (failed(readProp())) {
+      return failure();
+    }
+  }
```
iangneal: Extra scope here and below seems unnecessary

iangneal (changes requested):
Couple of quick comments, but it is overall looking good

include/llzk/Dialect/Shared/TypeConversionPatterns.h
tim-hoffman: That's actually the preferred approach for cases where the distinction is simple

iangneal (approved):
LGTM!

PR #411, opened 2026-04-15
Interval analysis precision update

iangneal:
- Fix precision of member writes when the written value is an array value
- Update modulus operator handling for intervals


lib/Analysis/Intervals.cpp
```diff
@@ -552,7 +552,25 @@ FailureOr<Interval> signedIntDiv(const Interval &lhs, const Interval &rhs) {
...
+    if (rhs.lhs() == f.zero()) {
+      return Interval::Entire(f);
+    }
+    return UnreducedInterval(f.zero(), rhs.lhs() - f.one()).reduce(f);
+  }
+
+  if (rhs.lhs() == f.zero()) {
+    return UnreducedInterval(f.zero(), rhs.b).reduce(f);
+  }
+  return UnreducedInterval(f.zero(), rhs.b - f.one()).reduce(f);
```
tim-hoffman: The switch from using `lhs()`/`rhs()` methods to direct field ref `b` threw me off for a second. Since we have locals named `lhs` and `rhs` I think it's clearer to just use the direct field refs (and that's what other Interval methods I scanned do as well).

lib/Analysis/Intervals.cpp
```diff
@@ -552,7 +552,25 @@ FailureOr<Interval> signedIntDiv(const Interval &lhs, const Interval &rhs) {
...
+  if (rhs.isDegenerate()) {
+    if (rhs.lhs() == f.zero()) {
+      return Interval::Entire(f);
+    }
+    return UnreducedInterval(f.zero(), rhs.lhs() - f.one()).reduce(f);
+  }
+
+  if (rhs.lhs() == f.zero()) {
+    return UnreducedInterval(f.zero(), rhs.b).reduce(f);
+  }
```
tim-hoffman: It seems off to me that this case which could have division by zero produces a smaller interval than the `rhs.isDegenerate()` case where it's a known division by zero.

tim-hoffman (approved):
LGTM

PR #412, opened 2026-04-16
Add basic RAM operation support

matthew-a-klein:
Add basic support for zk DSLs that use memory operations during witness generation, providing `load` and `store` ops. 
Both ops carry the `WitnessGen` trait and use direct `index` addressing with no allocation or region separation.


include/llzk-c/Dialect/RAM.h
```diff
@@ -0,0 +1,37 @@
+//===-- RAM.h - C API for RAM dialect ---------------------------*- C -*-===//
```
tim-hoffman: nit: line length
```suggestion
//===-- RAM.h - C API for RAM dialect -----------------------------*- C -*-===//
```

include/llzk-c/Dialect/RAM.h
```diff
@@ -0,0 +1,37 @@
...
+//
+// Part of the LLZK Project, under the Apache License v2.0.
+// See LICENSE.txt for license information.
+// Copyright 2025 Veridise Inc.
```
tim-hoffman: Let's use this copyright in all new files
```suggestion
// Copyright 2026 Project LLZK
```

iangneal: Nit:
```suggestion
// Copyright 2026 Project LLZK
```

include/llzk/Dialect/RAM/IR/Ops.td
```diff
@@ -0,0 +1,59 @@
+//===-- Ops.td ---------------------------------------------*- tablegen -*-===//
+//
+// Part of the LLZK Project, under the Apache License v2.0.
+// See LICENSE.txt for license information.
+// Copyright 2025 Veridise Inc.
```
iangneal: Nit:
```suggestion
// Copyright 2026 Project LLZK
```

include/llzk/Dialect/RAM/IR/Ops.td
```diff
@@ -0,0 +1,59 @@
...
+#define LLZK_RAM_OPS
+
+include "llzk/Dialect/Shared/Types.td"
+include "llzk/Dialect/RAM/IR/Dialect.td"
+include "llzk/Dialect/Function/IR/OpTraits.td"
+
+class RAMDialectOp<string mnemonic, list<Trait> traits = []>
+    : Op<RAMDialect, mnemonic, traits>;
+
+def LLZK_LoadOp : RAMDialectOp<"load", [WitnessGen]> {
```
iangneal: These ops should have memory effects traits (https://github.com/llvm/llvm-project/blob/f14644d085de8e835125b4260e3f3ceee8681f93/mlir/include/mlir/Interfaces/SideEffectInterfaces.td#L70-L88)

include/llzk/Dialect/RAM/IR/Ops.td
```diff
@@ -0,0 +1,59 @@
...
+
+    Example:
+
+    ```llzk
+    %val = ram.load %addr : !felt.type
+    ```
+  }];
+
+  let arguments = (ins Index:$addr);
+  let results = (outs AnyLLZKType:$val);
```
tim-hoffman: We need to further define the semantics here. Is this index a raw byte offset? Then we also need to specify the size of all LLZK types that can be stored in RAM. Currently, that's specified as any type which is probably too broad. All LLZK types are `I1, Index, String, Felt, Struct, Array, TypeVar, Pod`.

lib/Dialect/RAM/IR/Dialect.cpp
```diff
@@ -0,0 +1,30 @@
+//===-- Dialect.cpp - RAM dialect implementation -----------------*- C++ -*-===//
```
tim-hoffman: nit: line length
```suggestion
//===-- Dialect.cpp - RAM dialect implementation ----------------*- C++ -*-===//
```

tim-hoffman (commented):
Thanks for your contribution! Here are my initial thoughts.

iangneal (commented):
A handful of other small comments from me, apologizes for overlap between my comments and Tim's

include/llzk/Dialect/RAM/IR/Ops.td
matthew-a-klein: Have done the following: 
1)Restricted storage values to only allow Field elements. Callers who wish to store integere types need to widen first. 
2) Made explicit in the documentation that indeces are slot values, not byte offsets. Although now  the confusion should be slightly less because we can only store field elements. 
In an ideal world, from the noir -> LLZK perspective, we would be able to store integers of mutliple lengths, but the blow up of this would be far too large to justify. RIght now we are casting to Felt before storage.

include/llzk/Dialect/RAM/IR/Ops.td
matthew-a-klein: These now added

include/llzk/Dialect/RAM/IR/Ops.td
0xddom: Putting the responsibility of widening on the user requires that they have to emit the IR for the widening routine each time memory is touched. I can't speak for every user but I feel like that could be error prone or increase the size of the emitted IR with the widening logic. Also means that there isn't a standard widening logic for complex types (do we want one?). I think we could define a type interface where types define if they have a ram layout and how it looks like. That way passing a complex type that implements said interface to these ops would apply the operation to the whole address range taken by the type, keeping the IR more terse.

0xddom: One question, for your use case in Noir, do you want to read/write structs into RAM? If that's the case, manually widening the struct may be tricky since `struct.new` is restricted to the scope struct's type and other struct types are meant to be read from the result of a call to their `@compute` function.

matthew-a-klein: We only need to read and write atomic values

tim-hoffman (approved):
LGTM. Thank you @matthew-a-klein for your contribution!

PR #413, opened 2026-04-16
Convert `SourceRefAnalysis` to a sparse analysis

iangneal:
- `SourceRefAnalysis` is now a sparse forward analysis instead of a dense forward analysis. The semantics of the analysis remain unchanged
- Also fixes a soundness issue with the interval analysis for a niche loop yield case.


test/Analysis/interval_analysis/interval_analysis_pass_compute.llzk
```diff
@@ -542,12 +542,15 @@ module attributes {llzk.lang} {
       %0 = arith.constant 0 : index
       %1 = arith.constant 1 : index
       %f0 = felt.const 0
+      // COM: It seems since the loop always yields
+      // "3", but we actually don't know if the loop will run in practice (in which
+      // case it would yield "0"). We could get a tighter band but it's likely not
+      // worth optimizing for this case, since yielding a constant from a loop is
+      // not a useful use case, most likely.
```
tim-hoffman: nit: wording here is awk

lib/Analysis/ConstraintDependencyGraph.cpp
```diff
@@ -37,261 +37,257 @@ using namespace component;
...
-        updated |= after->setValue(result, *resultRef);
+  if (auto arrayAccessOp = llvm::dyn_cast<ArrayAccessOpInterface>(op)) {
+    if (llvm::isa<WriteArrayOp, InsertArrayOp>(arrayAccessOp)) {
+      auto array = arrayAccessOp.getArrRef();
+      auto it = operandVals.find(array);
+      ensure(it != operandVals.end(), "improperly constructed operandVals map");
+      const auto &currVals = it->second;
+
+      std::vector<SourceRefIndex> indices;
+      for (unsigned i = 0; i < arrayAccessOp.getIndices().size(); ++i) {
```
tim-hoffman: I would use `size_t` for `i` (since that's the return type of `arrayAccessOp.getIndices().size()` and the natural type for array indexing) and then add assertion before `arrayType.getDimSize(i)`: `assert(i <= std::numeric_limits<unsigned>::max()`. Bonus points if you make the cast explicit with `static_cast<unsigned>(i)`.

tim-hoffman: There are similar loops in `arraySubdivisionOpUpdate()` and `visitOperation()` below that should be updated as well.

iangneal: You had me at "bonus points."

test/Analysis/interval_analysis/interval_analysis_pass_compute.llzk
iangneal: Something shiny must have rolled across my desk when I was updating the comment.

lib/Analysis/ConstraintDependencyGraph.cpp
```diff
@@ -168,7 +169,7 @@ LogicalResult SourceRefAnalysis::visitOperation(
     }
 
     SourceRefLatticeValue newArrayVal(createArray.getType().getShape());
-    for (unsigned i = 0; i < elements.size(); i++) {
+    for (size_t i = 0; i < elements.size(); i++) {
       (void)newArrayVal.getElemFlatIdx(i).setValue(operandVals.at(elements[i])->getValue());
```
tim-hoffman, line 173: Can you add assert and cast for `getElemFlatIdx(unsigned)`

lib/Analysis/ConstraintDependencyGraph.cpp
```diff
@@ -220,7 +221,7 @@ void SourceRefAnalysis::visitExternalCall(
   const auto returnSites = predecessors->getKnownPredecessors();
 
   std::unordered_map<SourceRef, SourceRefLatticeValue, SourceRef::Hash> translation;
-  for (unsigned i = 0; i < funcOp.getNumArguments(); i++) {
+  for (size_t i = 0; i < funcOp.getNumArguments(); i++) {
```
tim-hoffman: Actually `getNumArguments()` returns `unsigned` so this one was correct before

tim-hoffman (commented):
A couple more things and it also looks like there's a merge conflict now

lib/Analysis/ConstraintDependencyGraph.cpp
iangneal: In looking at `getElemFlatIdx`, it should actually take a `size_t` instead of an `unsigned` as the input parameter, since it uses `vector::at` and `vector::size`, which are all `size_t` typed. So this is just downcasting to then upcast, which is a little silly. I'll fix it on the `getElemFlatIdx` side.

tim-hoffman: Good catch, I didn't even look further than its function signature.

tim-hoffman (approved):
LGTM

PR #419, opened 2026-04-21
Add getters and builders for TemplateParamOp and TemplateExprOp in TemplateBuilder

bikashxtech:
This PR adds getter and builder utilities for template-specific operations
TemplateParamOp and TemplateExprOp in TemplateBuilder.

Specifically:
- Added getParam(...)
- Added getExpr(...)
- Added insertParam(...)
- Added insertExpr(...)

These follow existing ModuleLikeBuilder conventions and use
TemplateOp::getConstNamed for lookup consistency.

### DIFF
--- a/changelogs/unreleased/add-templatebuilder-param-expr.yaml
+++ b/changelogs/unreleased/add-templatebuilder-param-expr.yaml
@@ -1,2 +1,2 @@
-category: Added
-summary: Add getters and builders for TemplateParamOp and TemplateExprOp in TemplateBuilder
+added:
+  - Add getters and builders for TemplateParamOp and TemplateExprOp in TemplateBuilder
--- a/changelogs/unreleased/add-templatebuilder-param-expr.yaml
+++ b/changelogs/unreleased/add-templatebuilder-param-expr.yaml
@@ -1,2 +1,3 @@
 added:
   - Add getters and builders for TemplateParamOp and TemplateExprOp in TemplateBuilder
+  
--- a/changelogs/unreleased/add-templatebuilder-param-expr.yaml
+++ b/changelogs/unreleased/add-templatebuilder-param-expr.yaml
@@ -1,0 +1,2 @@
+category: Added
+summary: Add getters and builders for TemplateParamOp and TemplateExprOp in TemplateBuilder
--- a/include/llzk/Dialect/Shared/Builders.h
+++ b/include/llzk/Dialect/Shared/Builders.h
@@ -310,7 +310,7 @@ public:
   /// Get the associated template of this builder.
   polymorphic::TemplateOp &getTemplate() { return myTemplate; }
 
-  mlir::FailureOr<polymorphic::TemplateParamOp> getParam(std::string_view name) const {
+  mlir::FailureOr<polymorphic::TemplateParamOp> getParam(std::string_view name) {
     auto op = myTemplate.getConstNamed<polymorphic::TemplateParamOp>(name);
     if (op) {
       return op;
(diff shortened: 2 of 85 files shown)


changelogs/unreleased/add-templatebuilder-param-expr.yaml
```diff
@@ -0,0 +1,2 @@
+added:
+  - Add getters and builders for TemplateParamOp and TemplateExprOp in TemplateBuilder
```
iangneal:
```suggestion
  - Add getters and builders for TemplateParamOp and TemplateExprOp in TemplateBuilder

```

include/llzk/Dialect/Shared/Builders.h
```diff
@@ -310,11 +310,66 @@ class TemplateBuilder : public ModuleLikeBuilder<TemplateBuilder> {
...
+    auto op = myTemplate.getConstNamed<polymorphic::TemplateExprOp>(name);
+    if (op) {
+      return op;
+    }
+    return mlir::failure();
+  }
+
+  TemplateBuilder &
+  insertParam(std::string_view name, mlir::Location loc, mlir::TypeAttr type = {}) {
+
```
iangneal, line 334:
```suggestion
```

include/llzk/Dialect/Shared/Builders.h
```diff
@@ -310,11 +310,66 @@ class TemplateBuilder : public ModuleLikeBuilder<TemplateBuilder> {
...
-  // TODO: other builders for template-specific ops like param/expr
+    return *this;
+  }
+
+  inline TemplateBuilder &insertParam(std::string_view name) {
+    return insertParam(name, getUnknownLoc());
+  }
+
+  TemplateBuilder &insertExpr(std::string_view name, mlir::Location loc) {
+
```
iangneal, line 359:
```suggestion
```

include/llzk/Dialect/Shared/Builders.h
```diff
@@ -310,11 +310,66 @@ class TemplateBuilder : public ModuleLikeBuilder<TemplateBuilder> {
...
+    auto &region = getBodyRegion();
+    if (region.empty()) {
+      region.emplaceBlock();
+    }
+
+    builder.setInsertionPointToEnd(&region.front());
+
+    auto nameAttr = builder.getStringAttr(name);
+
+    builder.create<polymorphic::TemplateParamOp>(loc, nameAttr, type);
```
iangneal, line 346: Are these create methods tied into the automatic symbol renaming logic or is this going to cause an error if you try to insert the same symbol twice?

iangneal (changes requested):
A few nits and a question

include/llzk/Dialect/Shared/Builders.h
bikashxtech: These insert methods currently follow the existing TemplateBuilder insertion conventions and rely on TemplateOp's symbol handling rather than performing automatic renaming themselves. So they behave consistently with the existing builder utilities duplicate symbol handling is delegated to the surrounding TemplateOp infrastructure rather than handled inside the insert helpers.If you'd prefer explicit duplicate-name checking or integration with a renaming utility here, I'm happy to update the implementation.

iangneal (changes requested):
Hey @bikashxtech, thank you for this contribution! Here's a couple things I'd like us to do here before we merge:

- `ModuleBuilder` has some unit tests in [unittests/IR/BuilderTests.cpp](https://github.com/project-llzk/llzk-lib/blob/main/unittests/IR/BuildersTests.cpp). Let's add some tests for the new methods to check that insert/lookup works as expected, and that multiple insertions of the same expression/parameter name yields a reasonable error.
- If repeated insertion of the same symbol name doesn't cause an error, then the builder itself should track symbol names: this is the strategy that `ModuleBuilder` and the `ModuleLikeBuilder` take, as implemented in the `ensure*Exists` and `ensureNoSuch*` methods.

bikashxtech:
@iangneal  I’ve added the requested unit tests and updated the builder to handle duplicate insertions with appropriate errors. Please let me know if there’s anything else I should improve.

iangneal (approved):
@bikashxtech looks good to us! Thanks again for your contribution!

PR #421, opened 2026-04-22
Add missing CAPI docs to clear up some of the `llzk-rs` build warnings

tim-hoffman:


iangneal (approved):

0xddom:
Note that the functions generated by `mlir-tblgen`, I think, don't include documentation so I think we should remove the warning from `llzk-sys` at some point because we are never going to fully remove the warnings

tim-hoffman:
> Note that the functions generated by `mlir-tblgen`, I think, don't include documentation so I think we should remove the warning from `llzk-sys` at some point because we are never going to fully remove the warnings

I noticed that too and was annoyed. I got it down to 10 warnings with https://github.com/project-llzk/llzk-rs/pull/147

PR #422, opened 2026-04-22
Fix -Wshadow error on SymbolUserHelper::benefit parameter

matthew-a-klein:
The `SymbolUserHelper` constructor parameter `benefit` (introduced in #378)
shadows the inherited `mlir::Pattern::benefit` member. GCC's `-Wshadow` fires
on this when MLIR headers are included via `-I` (non-system) rather than
`-isystem`, breaking downstream builds that consume LLZK as a library.

Rename the parameter to `patternBenefit` to avoid the collision. Updates the
two `/*benefit=*/` argument comments at `super(...)` call sites for
consistency; leaves the two direct calls to `OpConversionPattern<...>(...)`
untouched since MLIR's own parameter is named `benefit`.

No behavior change.

### DIFF
--- a/lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
+++ b/lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
@@ -219,10 +219,10 @@ private:
   const DenseMap<Attribute, Attribute> &paramNameToValue;
 
   SymbolUserHelper(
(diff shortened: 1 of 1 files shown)


0xddom (approved):
LGTM, thank you! The inconsistencies between compilers' flags sometimes create bizarre issues..

PR #423, opened 2026-04-22
fix: allow dead value removal to be used on `poly.expr`

tim-hoffman:
Removes the `Pure` trait from `poly.yield` op so the `remove-dead-values` pass can be used on `poly.expr` ops. With the `Pure` trait, `poly.yield` was not seen as a value sink for the liveness analysis which results in everything being removed expect the `poly.yield` itself (for some reason) which resulted in an invalid Value ref for its operand.

### DIFF
--- a/changelogs/unreleased/th__dead_value_cleanup_pass.yaml
+++ b/changelogs/unreleased/th__dead_value_cleanup_pass.yaml
@@ -1,0 +1,2 @@
+removed:
+  - "`Pure` trait from `poly.expr` operation (allows dead value removal to be used on `poly.expr`)"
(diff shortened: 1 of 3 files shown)


shankarapailoor (approved):
LGTM. There are some comments like “Replicate how it is done in…”. Do we need those or are those prompts for a coding agent?

tim-hoffman:
> LGTM. There are some comments like “Replicate how it is done in…”. Do we need those or are those prompts for a coding agent?

We should leave it. I was trying to reproduce an error I was seeing when I ran this on the circom frontend and my first attempt just used the simpler C++ API to setup the PassManager but it wasn't replicating the error in the same way so I reverse engineered the CAPI calls used from the circom side to reproduce the issue faithfully and wrote the comment to reflect that.

PR #425, opened 2026-04-23
Enable further simplification of `poly.expr` body

tim-hoffman:
Additions to allow `remove-dead-values` pass to further simplify `poly.expr` body:
  - `Pure` trait on `poly.read_const` op
  - `AlwaysSpeculatable` trait on `llzk.nondet` op (actually this was just unintentionally removed when `Pure` was dropped by a prior PR)
  - canonicalizer for `llzk.nondet` op to remove it when its result is not used

Fixes an assertion failure in `remove-dead-values` pass when there are empty `else` regions. Wraps the `remove-dead-values` pass with updates to `scf.if` ops to avoid the assertion failure. The bug is likely fixed in more recent versions of LLVM because https://github.com/llvm/llvm-project/blob/main/mlir/lib/Transforms/RemoveDeadValues.cpp is substantially different from LLVM 20, containing no calls to `Region::front()` which is where the assertion came from.

### DIFF
--- a/test/Dialect/LLZK_and_Builtin/nondet_preservation.llzk
+++ b/test/Dialect/LLZK_and_Builtin/nondet_preservation.llzk
@@ -1,5 +1,7 @@
 // RUN: llzk-opt -split-input-file --pass-pipeline='builtin.module(cse, canonicalize)' -verify-diagnostics %s | FileCheck --enable-var-scope %s
 
+// Ensures that `cse` and `canonicalize` passes do not combine multiple `llzk.nondet` operations because
+// adding a constraint to a nondet value makes it non-interchangeable with other nondet values.
 module attributes {llzk.lang} {
   struct.def @A {
     function.def @compute() -> !struct.type<@A> attributes {function.allow_witness} {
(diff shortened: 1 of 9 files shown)


test/FrontendLang/Circom/poly_expr_cleanup_2.llzk
```diff
@@ -0,0 +1,62 @@
...
+// CHECK-NEXT:        %[[VAL_7:[0-9a-zA-Z_\.]+]] = array.new %[[VAL_3]], %[[VAL_2]] : <2 x !felt.type<"bn128">>
+// CHECK-NEXT:        %[[VAL_8:[0-9a-zA-Z_\.]+]] = felt.add %[[VAL_6]], %[[VAL_1]] : !felt.type<"bn128">, !felt.type<"bn128">
+// CHECK-NEXT:        %[[VAL_9:[0-9a-zA-Z_\.]+]] = cast.toindex %[[VAL_0]] : !felt.type<"bn128">
+// CHECK-NEXT:        array.write %[[VAL_7]]{{\[}}%[[VAL_9]]] = %[[VAL_8]] : <2 x !felt.type<"bn128">>, !felt.type<"bn128">
+// CHECK-NEXT:        %[[VAL_10:[0-9a-zA-Z_\.]+]] = bool.cmp gt(%[[VAL_6]], %[[VAL_5]]) : !felt.type<"bn128">, !felt.type<"bn128">
+// CHECK-NEXT:        scf.if %[[VAL_10]] {
+// CHECK-NEXT:          %[[VAL_11:[0-9a-zA-Z_\.]+]] = cast.toindex %[[VAL_4]] : !felt.type<"bn128">
+// CHECK-NEXT:          %[[VAL_12:[0-9a-zA-Z_\.]+]] = array.read %[[VAL_7]]{{\[}}%[[VAL_11]]] : <2 x !felt.type<"bn128">>, !felt.type<"bn128">
+// CHECK-NEXT:          %[[VAL_13:[0-9a-zA-Z_\.]+]] = cast.toindex %[[VAL_0]] : !felt.type<"bn128">
+// CHECK-NEXT:          %[[VAL_14:[0-9a-zA-Z_\.]+]] = array.read %[[VAL_7]]{{\[}}%[[VAL_13]]] : <2 x !felt.type<"bn128">>, !felt.type<"bn128">
```
tim-hoffman, line 56: It would be nice to remove these dead reads as well, and then the write since there are no reads or uses of the Value aside from writes. I think we would need to define side effects for array ops and probably write a custom pass to remove them (the `remove-dead-values` pass only removes ops that "don't affect memory in any way" so maybe it would still be of some help if we used the more general "side effects" instead of "memory effects").

test/Dialect/LLZK_and_Builtin/nondet_preservation.llzk
```diff
@@ -1,5 +1,7 @@
 // RUN: llzk-opt -split-input-file --pass-pipeline='builtin.module(cse, canonicalize)' -verify-diagnostics %s | FileCheck --enable-var-scope %s
 
+// Ensures that `cse` and `canonicalize` passes do not combine multiple `llzk.nondet` operations because
+// adding a constraint to a nondet value makes it non-interchangeable with other nondet values.
```
iangneal, line 4: This comment would be good to add as part of the `llzk.nondet` documentation, to explain why it is `AlwaysSpeculatable` and not `Pure`.

test/FrontendLang/Circom/poly_expr_cleanup_1.llzk
```diff
@@ -0,0 +1,272 @@
+// RUN: llzk-opt -split-input-file --pass-pipeline='any(canonicalize,sccp,canonicalize,remove-dead-values,canonicalize)' %s 2>&1 | FileCheck --enable-var-scope %s
```
iangneal, line 1: "Hey I heard you like canonicalization, so I canonicalized your canonicalized canonicalization for you!"

test/FrontendLang/Circom/poly_expr_cleanup_2.llzk
iangneal: We have the `-llzk-duplicate-read-write-elim` pass which should work for arrays.

iangneal (approved):
Couple of suggestions/comments but otherwise LGTM

test/FrontendLang/Circom/poly_expr_cleanup_2.llzk
tim-hoffman: I just tried running that after `remove-dead-values` here and they were not removed. So maybe we can improve on that pass some to handle this.

iangneal: Okay; those passes haven't been touched in a LONG time, so I wouldn't be surprised if there were some issues with them. Maybe we can add a ticket for updating/improving them.

iangneal (approved):
LGTM!

PR #426, opened 2026-04-23
add new tests

tim-hoffman:


iangneal (commented):
What's the purpose of adding these tests?

tim-hoffman:
> What's the purpose of adding these tests?

Examples from brainstorming what Daniel is working on right now and I figured why not keep them. I guess I could actually add the full lit output check.

iangneal:
> > What's the purpose of adding these tests?
> 
> Examples from brainstorming what Daniel is working on right now and I figured why not keep them. I guess I could actually add the full lit output check.

That's reasonable. Yeah, maybe let's add full lit output or use them as tests for some other passes, perhaps.

iangneal (approved):

PR #432, opened 2026-05-05
add wildcard template parameter option

tim-hoffman:
I think this is the best solution to the issue of unknown call target result types in the circom frontend. It touches the least surface area of LLZK among the options discussed there.

### DIFF
--- a/changelogs/unreleased/th__wildcard_template_param.yaml
+++ b/changelogs/unreleased/th__wildcard_template_param.yaml
@@ -1,2 +1,2 @@
 added:
-  - Wildcard '?' option for template paramters to function.call
+  - Wildcard '?' option for template parameters to function.call
--- a/changelogs/unreleased/th__wildcard_template_param.yaml
+++ b/changelogs/unreleased/th__wildcard_template_param.yaml
@@ -1,0 +1,2 @@
+added:
+  - Wildcard '?' option for template paramters to function.call
--- a/include/llzk/Dialect/Function/IR/Ops.td
+++ b/include/llzk/Dialect/Function/IR/Ops.td
@@ -293,6 +293,15 @@ def CallOp : FunctionDialectOp<
     #M = affine_map<(i)[] -> (5*i+1)>
     %r = function.call @A::@compute(%x){(%i)} : (!felt.type) -> !struct.type<@A<[#M]>>
     ```
+
+    When the call targets a free function within a `poly.template` region, the optional
+    template parameter list can be used to instantiate all `poly.param` symbols within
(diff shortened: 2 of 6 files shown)


include/llzk/Dialect/Function/IR/Ops.td
```diff
@@ -293,6 +293,15 @@ def CallOp : FunctionDialectOp<
...
     %r = function.call @A::@compute(%x){(%i)} : (!felt.type) -> !struct.type<@A<[#M]>>
     ```
+
+    When the call targets a free function within a `poly.template` region, the optional
+    template parameter list can be used to instantiate all `poly.param` symbols within
+    the template. If all `poly.param` symbols are used within the function signature,
+    this can be elided. Otherwise, it is required to instantiate the function. The `?`
+    wildcard can be used for any `poly.param` with a `poly.tvar` type restriction, even
+    those that cannot be inferred from the function signature. The wildcard allows for
+    inferrence of the type within the function body itself during the flattening pass
```
iangneal:
```suggestion
    inference of the type within the function body itself during the flattening pass
```

changelogs/unreleased/th__wildcard_template_param.yaml
```diff
@@ -0,0 +1,2 @@
+added:
+  - Wildcard '?' option for template paramters to function.call
```
iangneal:
```suggestion
  - Wildcard '?' option for template parameters to function.call
```

test/Dialect/Poly/typevar_pass.llzk
```diff
@@ -931,3 +931,63 @@ module attributes {llzk.lang} {
...
+      %0 = felt.const  0 : <"bn128">
+      %1 = poly.unifiable_cast %0 : (!felt.type<"bn128">) -> !poly.tvar<@T_return>
+      function.return %1 : !poly.tvar<@T_return>
+    }
+  }
+
+  poly.template @synth {
+    poly.param @T_return : !poly.tvar<@T_return>
+    function.def @synth() {
+      // original call goes here
```
iangneal:
```suggestion
      %0 = function.call @f::@f() : () -> @T_return
```
I assume that's what "original call goes here" is referring to adding here

iangneal (changes requested):

changelogs/unreleased/th__wildcard_template_param.yaml
tim-hoffman: I feel like I make this exact typo all the time 🤪

iangneal (approved):

PR #438, opened 2026-05-08
Fix incorrect behavior in `pod.new` constructor

0xddom:
Fixes the incorrect construction of a `pod.new` op when is has both record initialization operands and map operands. 
The size of the first segment was not initialized to the length of the record initialization operands list, 
causing a validation error since it was expecting the default amount of 0 record initialization operands.

This issue triggered even if the map operands list was empty as long as the op was explicitly constructed with that constructor 
(i.e. by using the `pod::new_with_affine_init` factory function in the Rust bindings).


0xddom:
I'm testing downstream using this branch. I'll mark it as ready for review once I confirm the fix

tim-hoffman (approved):
LGTM

PR #440, opened 2026-05-12
Raghav/array to scalar attr fix

raghav198:
`{signal}` and `{column}` were swapped when generating scalar members from an array member


iangneal (changes requested):
Can we add a lit test for regression?

iangneal (approved):
Can we add a lit test for regression?

PR #442, opened 2026-05-12
Basic WhileToFor pass implementation

raghav198:
This identifies when an `scf.while` loop is shaped like a for loop and translates it. Useful because the Circom frontend always generates a while loop, but for loops are sometimes better for analysis.


iangneal (commented):
Should also add support for the `LoopBoundsAttr`, if it is present.

raghav198:
@codex review

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+  auto isRuntimeConstant = [&op](Value val) -> bool {
+    // The value can't come from a block argument owned by the while loop
+    if (auto blockArg = dyn_cast<BlockArgument>(val)) {
+      return blockArg.getParentBlock()->getParentOp() != op;
+    }
+    return true;
+  };
+
+  (void)isRuntimeConstant;
```
tim-hoffman: do we need this line?

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+#define DEBUG_TYPE "while-to-for"
+
+using namespace mlir;
+
+namespace llzk {
+
+using namespace boolean;
+using namespace felt;
+using namespace scf;
```
tim-hoffman: nit: prefer to open this namespace outside the `llzk` namespace. To me it reads like "llzk::scf" when inside.

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+  // Block argument index of the induction variable in the *before* block
+  std::optional<size_t> ivarIndexBefore;
+  // Block argument index of the induction variable in the *after* block
+  std::optional<size_t> ivarIndexAfter;
+
+  void reset() {
+    lb.reset();
+    ub.reset();
+    step.reset();
+  }
```
tim-hoffman: I haven't reviewed usage yet... but should this reset `ivarIndex*` also?

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+      if (arg == v) {
+        index.emplace(i);
+        break;
+      }
+    }
+  };
+
+  // Find which # block arg the ivar is in the before and after blocks
+  getBlockArgIndex(ivarBefore, op.getBeforeArguments(), info.ivarIndexBefore);
+  getBlockArgIndex(ivarBefore, op.getConditionOp().getArgs(), info.ivarIndexAfter);
```
tim-hoffman, line 106: nit: more clear to just use `getAfterArguments()`

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+  // If the yielded final value of the induction variable has any uses, we can't cleanly transform
+  // this to an scf.for (which doesn't explicitly yield its induction var) without doing some extra
+  // computation. Lets just conservatively bail out in that case
+  auto yieldedIVar = op->getResults().drop_front(*info.ivarIndexAfter).front();
+  if (!yieldedIVar.use_empty()) {
+    return ForOpInfo {};
+  }
+
+  Value ivarAfter = op.getAfterArguments().drop_front(*info.ivarIndexAfter).front();
```
tim-hoffman: opt: can move this line after the `if (op->hasAttr(LoopBoundsAttr::name))`

lib/Transforms/LLZKWhileToForPass.cpp
tim-hoffman: Also, could define the lambda to where it's used later in this function

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+  if (!info.ivarIndexBefore.has_value() || !info.ivarIndexAfter.has_value()) {
+    return info;
+  }
+
+  // If the yielded final value of the induction variable has any uses, we can't cleanly transform
+  // this to an scf.for (which doesn't explicitly yield its induction var) without doing some extra
+  // computation. Lets just conservatively bail out in that case
+  auto yieldedIVar = op->getResults().drop_front(*info.ivarIndexAfter).front();
+  if (!yieldedIVar.use_empty()) {
+    return ForOpInfo {};
```
tim-hoffman: `ForOpInfo::success()` ensures everything has a value (or else transformation is skipped) so it should be fine to return the existing `info` here rather than creating a new empty value, right?

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  Value ivarAfter = op.getAfterArguments().drop_front(*info.ivarIndexAfter).front();
+
+  // We need an induction variable anyway, but if the loop has {llzk.loopbounds} we can skip trying
+  // to parse the rest of the bounds and just materialize constants
+  if (op->hasAttr(LoopBoundsAttr::name)) {
+    auto bounds = op->getAttrOfType<LoopBoundsAttr>(LoopBoundsAttr::name);
+
+    IRRewriter rewriter {op->getContext()};
```
tim-hoffman: You're not utilizing the added functionality of a rewriter here so I prefer the simpler `OpBuilder` here.

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+    rewriter.setInsertionPoint(op);
+
+    // Make these constant felts for now; the actual for op builder will later clean it up
+    info.lb = rewriter
+                  .create<felt::FeltConstantOp>(
```
tim-hoffman: felt namespace is open so no need for prefix on these 3

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+  if (!isRuntimeConstant(*info.ub)) {
+    info.ub.reset();
+  }
+  if (info.step.has_value() && !isRuntimeConstant(*info.step)) {
+    info.step.reset();
+  }
+  return info;
+}
+
+static inline llvm::FailureOr<mlir::scf::ForOp>
```
tim-hoffman: mlir and scf namespaces are open so you can remove the prefix on a lot of types below

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+transformWhileToFor(mlir::scf::WhileOp op, ForOpInfo info, mlir::RewriterBase &rewriter) {
+  llzk::ensure(info.success(), "attempting to convert non-constant while loop");
+
+  rewriter.setInsertionPointAfter(op);
+  mlir::IRMapping mapping;
+
+  auto copyValue = [op, &rewriter, &mapping](mlir::Value val) -> mlir::Value {
```
tim-hoffman: Renaming to `copyIfNeeded` would be a bit more meaningful

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+    }
+    return rewriter.create<cast::FeltToIndexOp>(val.getLoc(), val).getResult();
+  };
+
+  // Emit a prelude setting up the loop bounds
+  auto lb = copyValue(*info.lb);
+  auto ub = copyValue(*info.ub);
+  auto step = copyValue(*info.step);
+
+  // Store the original type of the scf.while's induction var so we can cast back if necessary
```
tim-hoffman, line 233: comment seems outdated

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+  };
+
+  // scf.for bounds have to be `index`, so we might have to cast them here felt -> index before
+  // building the op, and cast them back index -> felt inside the body
+  auto toIndex = [&rewriter](mlir::Value val) -> mlir::Value {
+    if (!isa<FeltType>(val.getType())) {
+      return val;
+    }
+    return rewriter.create<cast::FeltToIndexOp>(val.getLoc(), val).getResult();
+  };
```
tim-hoffman, line 226: We should additionally add a runtime assertion for each that the felt value does not overflow the index type.

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+  auto inductionVar = forOp.getInductionVar();
+  if (isa<FeltType>(ivarType)) {
+    // If the induction var was a felt, we need to cast it back to felt in the scf.for body
+    // Note that this means the body of the scf.for might cast it back to index again anyway, but
+    // --canonicalize should fix that
+    inductionVar =
+        rewriter.create<cast::IntToFeltOp>(forOp.getLoc(), ivarType, inductionVar).getResult();
+  }
+
+  // // mapping.map(*info.ivar, forOp.getInductionVar()); // I don't think I need this anymore
```
tim-hoffman: remove commented code

lib/Transforms/LLZKWhileToForPass.cpp
```diff
@@ -0,0 +1,304 @@
...
+      mapping.map(whileBody->getArgument(i), inductionVar);
+      continue;
+    }
+    // map uses of the old "iter args" to the scf.for's actual iter_args
+    mapping.map(
+        whileBody->getArgument(i), forOp.getRegionIterArg(i > info.ivarIndexAfter ? i - 1 : i)
+    );
+  }
+
+  for (auto &bodyOp : *op.getAfterBody()) {
```
tim-hoffman:
```suggestion
  for (auto &bodyOp : *whileBody) {
```

tim-hoffman (commented):
comments on the code; scanning through the tests now

test/Transforms/WhileToFor/while_fail.llzk
```diff
@@ -0,0 +1,87 @@
+// RUN: llzk-opt --llzk-while-to-for --canonicalize -split-input-file %s | FileCheck --enable-var-scope %s
```
tim-hoffman: It could be useful for the pass to emit warnings or notes stating why certain loops could not be transformed and you verify those in the test with `-verify-diagnostics`

tim-hoffman:
@raghav198 could also add an example where the before and after blocks have different number of block arguments?

lib/Transforms/LLZKWhileToForPass.cpp
raghav198: Ah, yeah because it shouldn't have a lower bound set yet I guess

lib/Transforms/LLZKWhileToForPass.cpp
raghav198: It's semantically different; `getAfterArguments` returns the block arguments of the after block; here I'm looking for the SSA values being passed into the block so I can figure out which block arg represents the induction variable.

lib/Transforms/LLZKWhileToForPass.cpp
raghav198: "A runtime assertion" as in a "compile-time" assertion that causes the pass to fail, or do you mean generate an assertion in the IR? 

The former isn't possible here because the value I'm casting might not be a constant; I think this would be a case where we would want whoever is lowering the cast op later to check the cast if necessary.

lib/Transforms/LLZKWhileToForPass.cpp
raghav198: Ah, its referencing the `Type ivarType = ...` and then `if (isa<FeltType>(ivarType)) { /* emit index casts */ }` a couple lines down

lib/Transforms/LLZKWhileToForPass.cpp
tim-hoffman: Oh yeah, I had oversimplified in my understanding of `scf.while` values passed between blocks.

lib/Transforms/LLZKWhileToForPass.cpp
tim-hoffman: I mean the latter, generate something like this:
```
%max = felt.const 0xFFFFFFFFFFFFFFFF
%1 = bool.cmp lt(%val, %max)
bool.assert %1
```

tim-hoffman: I actually think `0xFFFFFFFFFFFFFFFF` is incorrect because index is signless, not unsigned. But, if cast op overflow semantics use "assert" as default, then nothing needs to be added/changed here.

tim-hoffman (approved):
LGTM

PR #444, opened 2026-05-12
fix llzk-tblgen bug involving variadic operands

tim-hoffman:
Previously generated code using a fixed offset of 1 for each variadic operand instead of consulting the `operandSegmentSizes` attribute.


tim-hoffman:
@codex review

iangneal (approved):

PR #446, opened 2026-05-12
Fix another tblgen bug by using reliable offset calculation

tim-hoffman:
Followup to https://github.com/project-llzk/llzk-lib/pull/444. There was still a scenario where variadic operands were not accounted for correctly. This PR changes all operand getter/setter impls to use the `getODSOperandIndexAndLength` function generated by `mlir-tblgen` to correctly compute operand offsets.


tim-hoffman:
@codex review

iangneal (approved):

PR #452, opened 2026-05-14
Fix `llzk-tblgen` handling for VariadicOfVariadic

tim-hoffman:
Another issue pointed out by copilot review of `llzk-tblgen`: the segment size attribute (e.g. `mapOpGroupSizes` in several LLZK IR ops) was not updated properly. The function signature did not convey enough information to update that attribute so this PR introduces fully separate handling for `VariadicOfVariadic` operands that accepts a `MlirValueRange*` parameter instead of just `MlirValue*` which the `Variadic` uses. 

Additionally, I realized the manual update of `operandSegmentSizes` was not necessary because `mlir-tblgen` generates `get*Mutable()` on the C++ side and the CAPI wrapper can just call into that to set the values and the segment sizes attribute.


tim-hoffman:
@codex review

iangneal (approved):

PR #461, opened 2026-05-17
Fix DynamicAPInt shifts over 64 bits

1sgtpepper:
## Summary

Fixes #392.

`po2` previously computed powers of two with `APSInt::get(1) << shiftAmt`, but `APSInt::get(1)` uses a fixed default width, so shift amounts above that width could assert inside LLVM.

This constructs the power-of-two value with `shiftAmt + 1` bits and sets the target bit directly. It also tightens the shift bound so the computed bit width cannot overflow.

## Testing

- `unittests/Analysis/LLZK_Analysis_Tests --gtest_filter="*DynamicAPIntShiftTest*"`
  - 12 tests passed
- `unittests/IR/LLZK_IR_Tests --gtest_filter="BabyBearFoldTest.Shl*:BabyBearFoldTest.Shr*"`
  - 8 tests passed

### DIFF
--- a/lib/Util/DynamicAPIntHelper.cpp
+++ b/lib/Util/DynamicAPIntHelper.cpp
@@ -21,11 +21,12 @@ using namespace llvm;
 using namespace std;
 
 static DynamicAPInt po2(const DynamicAPInt &e) {
-  // Ensure parameter is not negative and that it can be safely cast to unsigned.
+  // Ensure parameter is not negative and that shiftAmt + 1 fits in unsigned.
   assert(e >= 0);
-  assert(e <= std::numeric_limits<unsigned>::max() /* upcast from unsigned -> int64_t */);
+  assert(e < std::numeric_limits<unsigned>::max() /* upcast from unsigned -> int64_t */);
   unsigned shiftAmt = llzk::toAPSInt(e).getZExtValue();
-  APSInt p = APSInt::get(1) << shiftAmt;
+  APSInt p(shiftAmt + 1, /* isUnsigned */ true);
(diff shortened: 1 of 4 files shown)


1sgtpepper:
Hi @tim-hoffman, could you pls review this when you have a moment? Thanks for your help!

lib/Util/DynamicAPIntHelper.cpp
```diff
@@ -21,11 +21,12 @@ using namespace llvm;
 using namespace std;
 
 static DynamicAPInt po2(const DynamicAPInt &e) {
-  // Ensure parameter is not negative and that it can be safely cast to unsigned.
+  // Ensure parameter is not negative and that shiftAmt + 1 fits in unsigned.
   assert(e >= 0);
-  assert(e <= std::numeric_limits<unsigned>::max() /* upcast from unsigned -> int64_t */);
+  assert(e < std::numeric_limits<unsigned>::max() /* upcast from unsigned -> int64_t */);
```
iangneal:
```suggestion
  // APInt/APSInt bitwidth is limited to max unsigned bits, so must be strictly
  // less than the max to accommodate for the sign bit
  assert(e < std::numeric_limits<unsigned>::max());
```
Clarifying the comment because the actual reason for the assertion is not for an upcast.

iangneal (approved):
Thank you for your contribution! I just had one suggestion about updating a comment but otherwise LGTM!

1sgtpepper:
Thanks for the review, @iangneal!

iangneal (approved):
LGTM!

PR #462, opened 2026-05-17
Fix flattening of parameterized callees

1sgtpepper:
## Summary

Fixes `llzk-flatten` when an instantiated struct body calls through a template parameter, such as `@S::@compute`.

The flattening pass now rewrites parameterized callee roots in cloned instantiated bodies to the concrete struct symbol path. It also avoids lookup-based rewrite patterns while a callee is still symbolic inside the original template.

Fixes #388.

## Testing

- `git diff --check`
- `clang-format --dry-run --Werror lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp`

### DIFF
--- a/lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
+++ b/lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
@@ -1284,11 +1284,12 @@ public:
         FuncDefOp newFunc = callTgt.clone();
         convertCalleesInPlace(newFunc, paramNameToConcrete);
 
-        // Insert before body conversion so nested concrete callees verify from a root module.
+        // Insert before body conversion so nested concrete callees verify from the root module. Use
+        // the `SymbolTable::insert()` function so that the name will be made unique if necessary.
         symTables.getSymbolTable(newTemplate).insert(newFunc);
         symTables.getSymbolTable(parentModule).insert(newTemplate, Block::iterator(parentTemplate));
         if (failed(applyBodyConversions(newFunc))) {
-          std::string newFuncName = newFunc.getSymName().str();
+          StringRef newFuncName = newFunc.getSymName();
           LLVM_DEBUG(
               llvm::dbgs() << "[InstantiateFuncAtCallOp]   body conversion failed for "
                            << newFuncName << '\n'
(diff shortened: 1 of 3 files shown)


lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
```diff
@@ -1235,8 +1282,13 @@ class InstantiateFuncAtCallOp final : public OpRewritePattern<CallOp> {
...
         // Clone and partially convert the function (concretize only the concrete params).
         FuncDefOp newFunc = callTgt.clone();
+        convertCalleesInPlace(newFunc, paramNameToConcrete);
+
+        // Insert before body conversion so nested concrete callees verify from a root module.
+        symTables.getSymbolTable(newTemplate).insert(newFunc);
+        symTables.getSymbolTable(parentModule).insert(newTemplate, Block::iterator(parentTemplate));
         if (failed(applyBodyConversions(newFunc))) {
-          StringRef newFuncName = newFunc.getSymName();
+          std::string newFuncName = newFunc.getSymName().str();
```
iangneal: Any reason for the change from `StringRef` to `std::string`? If not we can revert to `StringRef`

iangneal (dismissed):
Overall LGTM, just one small request

lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
```diff
@@ -1235,8 +1282,13 @@ class InstantiateFuncAtCallOp final : public OpRewritePattern<CallOp> {
 
         // Clone and partially convert the function (concretize only the concrete params).
         FuncDefOp newFunc = callTgt.clone();
+        convertCalleesInPlace(newFunc, paramNameToConcrete);
+
+        // Insert before body conversion so nested concrete callees verify from a root module.
```
tim-hoffman:
```suggestion
        // Insert before body conversion so nested concrete callees verify from the root module. Use
        // the `SymbolTable::insert()` function so that the name will be made unique if necessary.
```

tim-hoffman (commented):
LGTM as well, just one request to keep the comment about inserting via symbol table

1sgtpepper:
Thanks both for the reviews. I restored the `SymbolTable::insert()` uniqueness note in the comment and reverted the temporary `std::string` back to `StringRef`.

tim-hoffman (approved):
LGTM, thanks for your contribution!

PR #465, opened 2026-05-20
Conservatively handle control-dependent interval writes

### LINKED ISSUE #317: Handle branch operation unsoundness in interval analysis
Consider the following:

```
  struct.def @YieldForConst {
    struct.member @a: !felt.type

    function.def @compute(%x: index) -> !struct.type<@YieldForConst> {
      %self = struct.new : !struct.type<@YieldForConst>
      %0 = arith.constant 0 : index
      %1 = arith.constant 1 : index
      %f0 = felt.const 0
      scf.for %i = %0 to %x step %1 {
        %3 = felt.const 3
        // COM: This is currently unsound, as it doesn't account for whether
        // or not the loop will run.
        struct.writem %self[@a] = %3 : !struct.type<@YieldForConst>, !felt.type
        scf.yield
      }
      function.return %self : !struct.type<@YieldForConst>
    }

    function.def @constrain(%self: !struct.type<@YieldForConst>, %x: index) {
      function.return
    }
  }
}

// CHECK-LABEL: @YieldForConst StructIntervals {
// CHECK-NEXT: compute {
// CHECK-NEXT:    %arg0 in Entire
// CHECK-NEXT:    %self[@a] in Degenerate(3)
// CHECK-NEXT:  }

```

Here the write to `@a` is considered absolute, even though the loop may not run.

1sgtpepper:
## Summary

Fixes #317.

Fixes interval-analysis unsoundness for writes inside SCF regions that may not execute.

Previously, writes recorded through `writeResults` were treated as unconditional even when the write occurred inside control flow such as `scf.for` or `scf.if`. That could make the analysis report overly precise results when the write path is skipped, such as a loop with zero iterations.

This patch makes those storage writes conservative by joining the written value with the prior/no-write state.

## Details

- Treat `struct.writem` and `array.write` updates under `scf.for` / `scf.if` as control-dependent writes.
- Preserve the prior/no-write storage state for those writes.
- Use the joined storage value when refining existing readers.
- Update interval-analysis checks for the reported `scf.for` case and the adjacent one-armed `scf.if` case.

## Testing

```sh
git diff --check
clang-format --dry-run --Werror include/llzk/Analysis/IntervalAnalysis.h lib/Analysis/IntervalAnalysis.cpp
```

Full lit suite is expected to be covered by CI.

### DIFF
--- a/lib/Analysis/IntervalAnalysis.cpp
+++ b/lib/Analysis/IntervalAnalysis.cpp
@@ -70,7 +70,7 @@ bool isInMaybeSkippedScfRegion(Operation *op) {
     // `writeResults` is a storage side-channel, not a path-sensitive lattice.
     // Writes nested under branch/loop control may not be absolute on every path through the
     // enclosing op, so keep the prior state instead of treating the nested write as unconditional.
-    if (llvm::isa<scf::ForOp, scf::IfOp>(parent)) {
+    if (llvm::isa<scf::ForOp, scf::IfOp, scf::WhileOp>(parent)) {
       return true;
     }
   }
--- a/lib/Analysis/IntervalAnalysis.cpp
+++ b/lib/Analysis/IntervalAnalysis.cpp
@@ -61,6 +61,22 @@ ExpressionValue refineReducedInterval(const ExpressionValue &expr, const Interva
   return refined;
 }
 
+bool isInMaybeSkippedScfRegion(Operation *op) {
+  for (Operation *parent = op->getParentOp(); parent != nullptr; parent = parent->getParentOp()) {
+    if (llvm::isa<FuncDefOp>(parent)) {
+      return false;
+    }
+
+    // `writeResults` is a storage side-channel, not a path-sensitive lattice.
+    // Writes nested under branch/loop control may not be absolute on every path through the
+    // enclosing op, so keep the prior state instead of treating the nested write as unconditional.
+    if (llvm::isa<scf::ForOp, scf::IfOp>(parent)) {
+      return true;
+    }
+  }
+  return false;
+}
+
 std::optional<UnreducedInterval> getBooleanUnreducedInterval(const Interval &interval) {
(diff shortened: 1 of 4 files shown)


lib/Analysis/IntervalAnalysis.cpp
```diff
@@ -61,6 +61,22 @@ ExpressionValue refineReducedInterval(const ExpressionValue &expr, const Interva
...
+bool isInMaybeSkippedScfRegion(Operation *op) {
+  for (Operation *parent = op->getParentOp(); parent != nullptr; parent = parent->getParentOp()) {
+    if (llvm::isa<FuncDefOp>(parent)) {
+      return false;
+    }
+
+    // `writeResults` is a storage side-channel, not a path-sensitive lattice.
+    // Writes nested under branch/loop control may not be absolute on every path through the
+    // enclosing op, so keep the prior state instead of treating the nested write as unconditional.
+    if (llvm::isa<scf::ForOp, scf::IfOp>(parent)) {
```
iangneal: This should also include while loops

test/Analysis/interval_analysis/interval_analysis_pass_compute.llzk
iangneal, line 1: Can you also add a test that uses a while loop?

iangneal (changes requested):
Thank you again for your contributions! I just have a couple of comments:

1sgtpepper:
Thanks for the review, @iangneal! I've updated the PR to include scf.while in the maybe-skipped region handling and added a regression test covering writes inside a while loop.

iangneal (approved):
Great, thank you! LGTM!

1sgtpepper:
Hi, just checking whether anything else is needed from my side on this one. It looks approved and CI is green, but I’m happy to adjust if there’s another merge requirement I missed.

iangneal:
@Kuhai9801 no, nothing needed to be done---I just didn't click merge! Thanks for the reminder.

PR #467, opened 2026-05-20
Fix uniform random for large primes (>64 bits)

tim-hoffman:
`toAPSInt(field.prime()).getZExtValue()` would assert for prime >64 bits


tim-hoffman:
@codex review

iangneal (approved):

tim-hoffman:
@iangneal turns out the codex report about a bug in `toDynamicAPInt(const APSInt&)` was correct, my test just didn't catch it initially because both values ended up going through that function and both being incorrectly converted, thus still equal.

tim-hoffman:
@codex review

iangneal (approved):

PR #470, opened 2026-05-20
Shankara/range optimized smt encoding

shankarapailoor:

PR #471, opened 2026-05-21
Optimize SMT lowering for direct field division equalities

1sgtpepper:
## Summary

Addresses #383 by adding a narrow SMT lowering fast path for direct equality constraints over `felt.div`.

When a `felt.div` result is used exactly once by `constrain.eq`, the lowering now avoids materializing an auxiliary SMT quotient value and emits the modular cross-multiplied assertion directly.

## Details

For a direct equality like:

```mlir
%q = felt.div %x, %y : !felt.type, !felt.type
constrain.eq %q, %t : !felt.type
```

the SMT lowering now emits:

```text
ite(%y == 0, %t mod p == 0, (%y * %t) mod p == %x mod p)
```

This preserves the current zero-divisor convention from the generic lowering while removing the fresh quotient symbol in the single-use equality case.

The existing generic `felt.div` lowering remains unchanged for all other cases, including multi-use quotients.

## Tests

Added FileCheck coverage for:

- direct `%quotient == %target`
- flipped equality `%x == %quotient`
- multi-use quotient fallback to the existing generic lowering

### DIFF
--- a/backends/smt/lib/Conversions/SMTLoweringPass.cpp
+++ b/backends/smt/lib/Conversions/SMTLoweringPass.cpp
@@ -140,9 +140,66 @@ public:
   FeltDivConverter(TypeConverter &_typeConverter, MLIRContext *context, APSInt _prime)
       : OpConversionPattern {_typeConverter, context, /*benefit=*/2}, prime {std::move(_prime)} {}
 
+private:
+  LogicalResult rewriteSingleUseEquality(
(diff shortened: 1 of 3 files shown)


shankarapailoor:
Thanks for the contribution, but this will be subsumed by https://github.com/project-llzk/llzk-lib/pull/470/ as you can see here: https://github.com/project-llzk/llzk-lib/pull/470/changes#diff-5f11ae5e3b8619068a17a3bf25338728a8c46287de2b7d7e51adb1eb71da7138

Issue #136, opened 2025-06-09
Make interval analysis interprocedural


iangneal:
Currently, the interval analysis is intraprocedural. We should leverage subcomponent analyses (as we do for the constraint dependency graph analysis) to perform an interprocedural version of this analysis.

PR #475, opened 2026-05-22
Make interval analysis interprocedural for subcomponent constrains

1sgtpepper:
## Summary

Addresses #136 by allowing interval analysis to reuse concrete subcomponent constrain summaries.

This updates interval analysis to:
- propagate callee struct constrain intervals into parent-visible refs
- translate child `SourceRef`s through call operands before merging
- merge only when translation resolves to one concrete parent ref
- conservatively skip unresolved/template calls, dynamic multi-ref translations, and recursive constrain cycles

The patch also adds regression coverage for direct subcomponent propagation, static vs dynamic array paths, unresolved template calls, and mutual constrain-call cycles.

## Testing

- `git diff --check`
- `nix --extra-experimental-features "nix-command flakes" build -L --extra-substituters https://veridise-public.cachix.org --extra-trusted-public-keys "veridise-public.cachix.org-1:FvpZ8GzAj1mmJA5PnO9UgKxC6CQdmPutuIKtEpGmeig=" .#debugClang`

Result:
- Lit: 201 passed, 6 unsupported, 2 expected failures, 0 failed
- Gtests: 1086/1086 passed


test/Analysis/interval_analysis/interprocedural_subcomponent.llzk
```diff
@@ -0,0 +1,221 @@
...
+
+// CHECK-LABEL: @CycleA StructIntervals {
+// CHECK-DAG:     %arg0[@out] in Entire
+// CHECK-DAG:     %arg1[@out] in Entire
+// CHECK: }
+
+// CHECK-LABEL: @CycleB StructIntervals {
+// CHECK-DAG:     %arg0[@out] in Entire
+// CHECK-DAG:     %arg1[@out] in Entire
+// CHECK: }
```
iangneal, line 221:
```suggestion
// CHECK: }

// -----

module attributes {llzk.lang} {
  struct.def @Child {
    struct.member @out : !felt.type {llzk.pub}

    function.def @compute() -> !struct.type<@Child> {
      %self = struct.new : !struct.type<@Child>
      function.return %self : !struct.type<@Child>
    }

    function.def @constrain(%self: !struct.type<@Child>) {
      %out = struct.readm %self[@out] : !struct.type<@Child>, !felt.type
      %seven = felt.const 7
      %res = bool.cmp le(%out, %seven)
      %true = arith.constant true
      constrain.eq %res, %true : i1
      function.return
    }
  }

  struct.def @Parent {
    struct.member @child : !struct.type<@Child>

    function.def @compute() -> !struct.type<@Parent> {
      %self = struct.new : !struct.type<@Parent>
      function.return %self : !struct.type<@Parent>
    }

    function.def @constrain(%self: !struct.type<@Parent>) {
      %child = struct.readm %self[@child] : !struct.type<@Parent>, !struct.type<@Child>
      function.call @Child::@constrain(%child) : (!struct.type<@Child>) -> ()

      %out = struct.readm %child[@out] : !struct.type<@Child>, !felt.type
      %one = felt.const 1
      %res = bool.cmp ge(%out, %one)
      %true = arith.constant true
      constrain.eq %res, %true : i1
      function.return
    }
  }
}

// CHECK-LABEL: @Child StructIntervals {
// CHECK-NEXT:     %arg0[@out] in TypeA:[ 0, 7 ]
// CHECK-NEXT: }

// CHECK-LABEL: @Parent StructIntervals {
// CHECK-NEXT:     %arg0[@child][@out] in TypeA:[ 1, 7 ]
// CHECK-NEXT: }
```

Just wanted to add some tests where additional constraints are added in the parent so we can show the parent intervals don't influence the child results.

test/Analysis/interval_analysis/interprocedural_subcomponent.llzk
iangneal, line 1: Here's a test case which I believe demonstrates a bug in the interprocedural logic:

```mlir
  struct.def @Child {
    struct.member @out : !felt.type {llzk.pub}

    function.def @compute(%in : !felt.type) -> !struct.type<@Child> {
    }

    function.def @constrain(%self: !struct.type<@Child>, %in : !felt.type) {
      %out = struct.readm %self[@out] : !struct.type<@Child>, !felt.type
      constrain.eq %out, %in : !felt.type
      %seven = felt.const 7
      %res = bool.cmp le(%out, %seven)
      %true = arith.constant true
      constrain.eq %res, %true : i1
      function.return
    }
  }

  struct.def @Parent {
    struct.member @child_in : !felt.type
    struct.member @child : !struct.type<@Child>

    function.def @compute() -> !struct.type<@Parent> {
      %self = struct.new : !struct.type<@Parent>
      function.return %self : !struct.type<@Parent>
    }

    function.def @constrain(%self: !struct.type<@Parent>) {
      %in = struct.readm %self[@child_in] : !struct.type<@Parent>, !felt.type
      %three = felt.const 3
      %res = bool.cmp ge(%in, %three)
      %true = arith.constant true
      constrain.eq %res, %true : i1

      %child = struct.readm %self[@child] : !struct.type<@Parent>, !struct.type<@Child>
      // input is [3, prime - 1] before the call
      function.call @Child::@constrain(%child, %in) : (!struct.type<@Child>, !felt.type) -> ()
      // with child restrictions, output and input should be [3, 7]

      function.return
    }
  }
}

// CHECK-LABEL: @Child StructIntervals {
// CHECK-NEXT:     %arg0[@out] in TypeA:[ 0, 7 ]
// CHECK-NEXT:     %arg1 in TypeA:[ 0, 7 ]
// CHECK-NEXT: }

// CHECK-LABEL: @Parent StructIntervals {
// CHECK-NEXT:     %arg0[@child_in] in TypeA:[ 3, 7 ]
// CHECK-NEXT:     %arg0[@child][@out] in TypeA:[ 3, 7 ]
// CHECK-NEXT: }
```

Expected output is:
```
@Child StructIntervals { 
    %arg0[@out] in TypeA:[ 0, 7 ]
    %arg1 in TypeA:[ 0, 7 ]
}
@Parent StructIntervals { 
    %arg0[@child_in] in TypeA:[ 3, 7 ]
    %arg0[@child][@out] in TypeA:[ 3, 7 ]
}
```

Actual output is:
```
@Child StructIntervals { 
    %arg0[@out] in TypeA:[ 0, 7 ]
    %arg1 in TypeA:[ 0, 7 ]
}
@Parent StructIntervals { 
    %arg0[@child_in] in TypeA:[ 3, 7 ]
    %arg0[@child][@out] in TypeA:[ 0, 7 ]
}
```

lib/Analysis/IntervalAnalysis.cpp
```diff
@@ -35,6 +37,18 @@ using namespace function;
 
 namespace {
 
+bool isOperationLive(DataFlowSolver &solver, Operation *op) {
```
iangneal: This could be moved into `AnalysisUtil.{h,cpp}`

iangneal (changes requested):
Hey @Kuhai9801! Thanks for the contribution. I think I've found a test case which exposes some issues---let me know what you think.

1sgtpepper:
Thanks again for the careful review. I pushed an update that should address the points you raised:

- Moved the shared `isOperationLive` helper into `AnalysisUtil`.
- Added the parent/child regression where the child keeps its local `[0, 7]` interval, while the parent refines the child output to `[1, 7]` after an additional parent-side constraint.
- Fixed the interprocedural case where a child equality constraint between `out` and an identity-passed input lets the caller refine both the input and child output to `[3, 7]`.
- Added a non-identity regression for `felt.neg` so the propagation stays limited to identity source refs and does not treat transformed operands as equivalent.

CI is now passing on the updated head, including the GCC/Clang test jobs and both debug/release Nix checks.

include/llzk/Analysis/AnalysisUtil.h
```diff
@@ -11,6 +11,10 @@
 
 #include <mlir/Analysis/DataFlowFramework.h>
 
+namespace mlir {
+class Operation;
+} // namespace mlir
+
```
iangneal: This is unnecessary given the `<mlir/Analysis/DataFlowFramework.h>` include.
```suggestion
```

test/Analysis/interval_analysis/interprocedural_subcomponent.llzk
```diff
@@ -0,0 +1,385 @@
...
+
+// CHECK-LABEL: @Child StructIntervals {
+// CHECK-NEXT:     %arg0[@out] in TypeA:[ 0, 7 ]
+// CHECK-NEXT:     %arg1 in TypeA:[ 0, 7 ]
+// CHECK-NEXT: }
+
+// CHECK-LABEL: @Parent StructIntervals {
+// CHECK-NEXT:     %arg0[@child_in] in TypeC:[ 3, 21888242871839275222246405745257275088696311157297823662689037894645226208582 ]
+// CHECK-NEXT:     %arg0[@child][@out] in TypeA:[ 0, 7 ]
+// CHECK-NEXT: }
```
iangneal, line 385: This seems wrong; `@Child::@constrain` constrains `%arg0[@child_in]` to be equal to `%arg0[@child][@out]`, so they should end up with the same interval, even if that interval is `Empty`.

iangneal (changes requested):

test/Analysis/interval_analysis/interprocedural_subcomponent.llzk
1sgtpepper: Thanks for flagging this. I think this case is intended to be different from the identity-input regression above.

Here the caller passes `%neg`, not `%in`, into `@Child::@constrain`:

```mlir
%neg = felt.neg %in : !felt.type
function.call @Child::@constrain(%child, %neg) : (!struct.type<@Child>, !felt.type) -> ()
```

So the callee equality relates `%arg0[@child][@out]` to the transformed SSA value `%neg`, not directly to `%arg0[@child_in]`. The interprocedural translation in this PR is intentionally limited to identity `SourceRef` operands, so I don’t think this test should merge the parent `child_in` interval with the child output interval.

A stronger future rule could propagate through `felt.neg` and refine `child_in` via inverse negation, but that would still produce a negated range rather than the same interval as `child.out`. Does that match your intended requirement, or would you prefer this PR to handle expression-level inverse propagation too?

iangneal: Yeah you're right on the equality comparison there, `%self[@child][@out]` isn't going to be equal to `%self[@child_in]` because of that negation. 

I do think we're still missing some interprocedural propagation, though. By looking at the SSA intervals (you can see these results via `-llzk-print-interval-analysis="print-ssa-intervals"`:

```
@Child StructIntervals { 
    %arg0[@out] in TypeA:[ 0, 7 ]
    %arg1 in TypeA:[ 0, 7 ]
}
@Child SSAIntervals {
    constrain {
        %arg1 in TypeA:[ 0, 7 ]
        %0 [struct.readm] in TypeA:[ 0, 7 ]
        %felt_const_7 [felt.const] in Degenerate(7)
        %1 [bool.cmp] in Degenerate(1)
        %true [arith.constant] in Degenerate(1)
    }
}
@Parent StructIntervals { 
    %arg0[@child_in] in TypeC:[ 3, 21888242871839275222246405745257275088696311157297823662689037894645226208582 ]
    %arg0[@child][@out] in Entire
}
@Parent SSAIntervals {
    constrain {
        %0 [struct.readm] in TypeC:[ 3, 21888242871839275222246405745257275088696311157297823662689037894645226208582 ]
        %felt_const_3 [felt.const] in Degenerate(3)
        %1 [bool.cmp] in Degenerate(1)
        %true [arith.constant] in Degenerate(1)
        %2 [felt.neg] in TypeC:[ 1, 21888242871839275222246405745257275088696311157297823662689037894645226208580 ]
    }
}
```

`%neg` is `TypeC:[ 1, 21888242871839275222246405745257275088696311157297823662689037894645226208580 ]`, so when that is passed in to the `@Child::@constrain` call, the result should be that `%self[@child][@out]` will be `TypeC:[ 1, 21888242871839275222246405745257275088696311157297823662689037894645226208580 ]`  further constrained to be less than 7 (so, `[1, 7]`) rather than `[0, 7]`. 

I'm less concerned with the backwards propagation of that constraint back to `@child_in`, but more wanting the input interval to be propagated into the `@Child::@constrain` call. Does that make sense?

1sgtpepper: Thanks @iangneal, that makes sense. I updated the interprocedural merge to include scalar call-site operand intervals in the callee direct-equality context.

For this case, `%neg` is still not treated as identity-equivalent to `%child_in`, so we don’t back-propagate through `felt.neg`. But its SSA interval is now intersected into the callee equality group for `%arg1 == %self[@out]`, and only the translated parent-visible child ref is merged back into the parent summary. That refines `%self[@child][@out]` to `[1, 7]` while leaving `%self[@child_in]` at its existing negation-derived interval.

I also updated the regression expectation for the non-identity operand case from `[0, 7]` to `[1, 7]`.

iangneal:
@codex review

lib/Analysis/IntervalAnalysis.cpp
```diff
@@ -1630,6 +1816,159 @@ LogicalResult StructIntervals::computeIntervals(
...
+          llvm::MapVector<SourceRef, Interval> translatedEqRefs;
+          Interval contextualInterval = Interval::Entire(ctx.getField());
+          bool hasInterval = false;
+          bool ambiguousTranslation = false;
+
+          for (auto memberIt = directEqRefs.member_begin(leaderIt);
+               memberIt != directEqRefs.member_end(); ++memberIt) {
+            Interval memberInterval = Interval::Entire(ctx.getField());
+            if (auto childIntervalIt = childIntervals.getConstrainIntervals().find(*memberIt);
+                childIntervalIt != childIntervals.getConstrainIntervals().end()) {
```
iangneal:
```suggestion
            const auto &ci = childIntervals.getConstrainIntervals();
            if (auto childIntervalIt = ci.find(*memberIt); childIntervalIt != ci.end()) {
```

lib/Analysis/IntervalAnalysis.cpp
```diff
@@ -66,6 +70,157 @@ std::optional<UnreducedInterval> getBooleanUnreducedInterval(const Interval &int
...
+    if (lhsRef.isConstant() || rhsRef.isConstant()) {
+      return;
+    }
+    eqRefs.unionSets(lhsRef, rhsRef);
+  });
+  return eqRefs;
+}
+
+FailureOr<SymbolLookupResult<FuncDefOp>>
+resolveCallableSilently(SymbolTableCollection &tables, CallOp fnCall) {
```
iangneal: This could be moved to the symbol lookup helper utilities, ideally also with a documentation comment.

lib/Analysis/IntervalAnalysis.cpp
```diff
@@ -1630,6 +1816,159 @@ LogicalResult StructIntervals::computeIntervals(
...
+        const StructIntervals &childIntervals = childAnalysis.getResult(ctx);
+        for (const auto &[childRef, childInterval] : childIntervals.getConstrainIntervals()) {
+          auto translatedRefs = translateRef(childRef, identityTranslations);
+          if (failed(translatedRefs)) {
+            continue;
+          }
+
+          std::optional<UnreducedInterval> childUnreduced = std::nullopt;
+          auto childUnreducedIt = childIntervals.getConstrainUnreducedIntervals().find(childRef);
+          if (childUnreducedIt != childIntervals.getConstrainUnreducedIntervals().end()) {
```
iangneal:
```suggestion
          const auto &cui = childIntervals.getConstrainUnreducedIntervals();
          if (auto childUnreducedIt = cui.find(childRef); childUnreducedIt != cui.end()) {
```

lib/Analysis/IntervalAnalysis.cpp
```diff
@@ -1630,6 +1816,159 @@ LogicalResult StructIntervals::computeIntervals(
       }
     }
 
+    // Child constrain calls refine parent-visible storage, but only after the callee
+    // summary is translated through the call operands. If translation cannot prove
+    // which parent ref owns a child interval, the local overapproximation remains.
+    if (fn.isStructConstrain()) {
+      auto mergeChildConstrainIntervals = [&](CallOp fnCall) {
```
iangneal, line 1799: This lambda is long enough that it would be nice to have some comments internally to break down what is going on and why.

iangneal (changes requested):
Overall LGTM, just a few minor requests before we merge:

iangneal (approved):
LGTM!

PR #477, opened 2026-05-22
Add `verif` dialect and ops

### LINKED ISSUE #434: Add `verif` dialect

iangneal:
- Added `verif.contract` for defining specs on structs and functions
- Added `verif.include` to invoke contracts from within contracts
- Added `verif.require_compute` and `verif.require_constrain` for encoding preconditions
- Added `verif.ensure_compute` and `verif.ensure_constrain` for encoding postconditions

Fixes https://github.com/project-llzk/llzk-lib/issues/434


iangneal:
@codex review

iangneal:
@codex review

iangneal:
@0xddom and @raghav198 would appreciate a review on this this week.

include/llzk/Dialect/Verif/IR/Ops.td
```diff
@@ -0,0 +1,398 @@
...
+
+def EnsureConstrainOp : EnsureOpBase<"ensure_constrain"> {
+  let summary = "constraint generation postcondition";
+}
+
+//===------------------------------------------------------------------===//
+// ContractDefOp
+//===------------------------------------------------------------------===//
+
+// TODO: Build methods directly from FuncDefOp and StructDefOp
```
raghav198: Can this TODO be more descriptive, or do we have an issue linked for it already?

test/Dialect/Verif/contracts_fail.llzk
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// -----
+
+module attributes {llzk.lang} {
+  function.def @check(%x: index) {
+    function.return
+  }
+
+  // expected-error@+1 {{'verif.contract' op expected a valid LLZK type but found 'i32'}}
```
raghav198: The targets of a `verif.contract` can only take/return valid LLZK types anyway, right? So is it enough to just check for a type mismatch here instead?

include/llzk/Dialect/Verif/IR/Ops.td
iangneal: Oh, it's already resolved, so I'll remove it.

raghav198 (approved):
LGTM

iangneal:
I'll wait to merge this since @0xddom mentioned he was working on a review for this as well.

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+  MLIRContext *ctx = odsBuilder.getContext();
+  SymbolRefAttr targetAttr = SymbolRefAttr::get(ctx, target);
+  // Find the target of the contract
+  FailureOr<SymbolLookupResultUntyped> targetRes =
+      lookupTopLevelSymbol(tables, targetAttr, odsBuilder.getBlock()->getParentOp());
+  if (failed(targetRes)) {
+    InFlightDiagnosticWrapper err(odsState.location);
+    err.append("could not find target \"@", targetAttr, "\"");
+    err.report();
+    return;
```
0xddom, line 170: This kind of checks should be done in the verification method. In the build method we should just take the data at face value and then check it in the other method

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+// operations. We'll verify contract-specific restrictions in `verify`.
+ParseResult ContractOp::parse(OpAsmParser &parser, OperationState &result) {
+  auto buildFuncType = [](Builder &builder, ArrayRef<Type> argTypes, ArrayRef<Type> results,
+                          function_interface_impl::VariadicFlag,
+                          std::string &) { return builder.getFunctionType(argTypes, results); };
+
+  return parseContractOp(
+      parser, result, getFunctionTypeAttrName(result.name), buildFuncType,
+      getArgAttrsAttrName(result.name)
+  );
```
0xddom: Why do we need this lambda here? We control the parse method

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+      );
+      return WalkResult::interrupt();
+    }
+    return WalkResult::advance();
+  });
+  if (res.wasInterrupted()) {
+    return failure();
+  }
+
+  return success();
```
0xddom:
```suggestion
  return failure(res.wasInterrupted());
```

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+  assert(!isNullOrEmpty(callParams) && "pre-condition");
+  assert((callParams.size() == llvm::range_size(targetParamDefs)) && "pre-condition");
+
+  for (auto [paramOp, attr] : llvm::zip_equal(targetParamDefs, callParams.getValue())) {
+    // Skip wildcards (`?` / kDynamic) - their value will be resolved by a later inference pass.
+    if (auto intAttr = llvm::dyn_cast<IntegerAttr>(attr)) {
+      if (isDynamic(intAttr)) {
+        continue;
+      }
+    }
```
0xddom:
```suggestion
    if (isDynamic(llvm::dyn_cast<IntegerAttr>(attr))) {
      continue;
    }
```

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+    auto it = unifications.find({FlatSymbolRefAttr::get(paramOp.getNameAttr()), Side::RHS});
+    if (it != unifications.end() && !typeParamsUnify({attr}, {it->second})) {
+      // Tested in call_with_template_params_fail.llzk
```
0xddom: There's no need for this comment

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+
+protected:
+  IncludeOp *includeOp;
+
+  virtual LogicalResult verifyInputs() = 0;
+  virtual LogicalResult verifyTemplateParams() = 0;
+
+  LogicalResult verifyNoTemplateInstantiations() {
+    if (!isNullOrEmpty(includeOp->getTemplateParamsAttr())) {
+      // Tested in call_with_template_params_fail.llzk
```
0xddom: Useless comment too

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+        llzk::getSymbolsUsedIn(tgtType.getInputs(), referencedInSignature);
+        llzk::getSymbolsUsedIn(tgtType.getResults(), referencedInSignature);
+
+        bool allParamsReferenced = llvm::all_of(realParams, [&](TemplateParamOp p) {
+          return referencedInSignature.contains(FlatSymbolRefAttr::get(p.getNameAttr()));
+        });
+        if (allParamsReferenced) {
+          return success();
+        }
+        // Tested in call_with_template_params_fail.llzk
```
0xddom: Useless comment

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+      if (failed(llzk::forceIntAttrTypes(callParams.getValue(), [this] {
+        return llzk::InFlightDiagnosticWrapper(this->includeOp->emitOpError());
+      }))) {
+        return failure();
+      }
+
+      // The instantiation list is present. Check it has exactly one entry per template param.
+      size_t numTemplateParams = llvm::range_size(realParams);
+      if (callParams.size() != numTemplateParams) {
+        // Tested in call_with_template_params_fail.llzk
```
0xddom: Useless comment

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+  // If the callee references a parameter of the template where this call appears, perform
+  // the subset of checks that can be done even though the target is unknown.
+  if (calleeAttr.getNestedReferences().size() == 1) {
+    if (TemplateOp parent = getParentOfType<TemplateOp>(*this)) {
+      if (parent.hasConstNamed<TemplateParamOp>(calleeAttr.getRootReference())) {
+        return this->emitError("expected parameterized callee to target a struct function")
+            .append(
+                " (i.e. \"@", FUNC_NAME_PRODUCT, "\", \"@", FUNC_NAME_COMPUTE, "\", or \"@",
+                FUNC_NAME_CONSTRAIN, "\")"
+            );
```
0xddom, line 788: This second message should probably be an attached note instead of part of the error message

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+  return FunctionType::get(getContext(), getArgOperands().getTypes(), /*results*/ {});
+}
+
+FailureOr<UnificationMap> IncludeOp::unifyTypeSignature(FunctionType other) {
+  UnificationMap unifications;
+  if (functionTypesUnify(getTypeSignature(), other, {}, &unifications)) {
+    return unifications;
+  } else {
+    return failure();
+  }
```
0xddom:
```suggestion
  }
  return failure();
```

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+}
+
+Value IncludeOp::getSelfValue() {
+  SymbolTableCollection tables;
+  auto callee = getCalleeTarget(tables);
+  assert(succeeded(callee) && "include callee must resolve");
+  if (!callee->get().hasStructTarget()) {
+    return nullptr;
+  }
+  assert(getNumOperands() > 0 && "include op must have a self operand");
```
0xddom, line 838: Even when the contract's target is a function?

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+void IncludeOp::setCalleeFromCallable(CallInterfaceCallable callee) {
+  setCalleeAttr(llvm::cast<SymbolRefAttr>(callee));
+}
+
+SmallVector<ValueRange> IncludeOp::toVectorOfValueRange(OperandRangeRange input) {
+  llvm::SmallVector<ValueRange, 4> output;
+  output.reserve(input.size());
+  for (OperandRange r : input) {
+    output.push_back(r);
+  }
```
0xddom: `SmallVector` doesn't have some kind of `insert` method that accepts an iterator?

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -0,0 +1,909 @@
...
+  return output;
+}
+
+Operation *IncludeOp::resolveCallableInTable(SymbolTableCollection *symbolTable) {
+  FailureOr<SymbolLookupResult<ContractOp>> res =
+      llzk::resolveCallable<ContractOp>(*symbolTable, *this);
+  if (failed(res) || res->isManaged()) {
+    // Cannot return pointer to a managed Operation since it would cause memory errors.
```
0xddom: That's why we have the lookup result class that handles memory of modules loaded on the fly once the result is no longer needed

unittests/CAPI/Dialect/Verif.cpp
```diff
@@ -0,0 +1,257 @@
...
+  );
+  wrapper.getBody().emplaceBlock();
+  unwrap(builder)->setInsertionPointToStart(&wrapper.getBody().front());
+
+  MlirOperation includeOp = llzkVerif_IncludeOpBuild(
+      builder, location, createFlatSymbolRefAttr(context, "Base"), {.values = nullptr, .size = 0},
+      mlirAttributeGetNull()
+  );
+
+  EXPECT_NE(includeOp.ptr, (void *)NULL);
```
0xddom: The checks below should crash if this pointer is null, so I think this check is a bit redundant

unittests/CAPI/Dialect/Verif.cpp
```diff
@@ -0,0 +1,257 @@
...
+      unwrap(location), name, mlir::SymbolRefAttr::get(unwrap(builder)->getContext(), target),
+      mlir::FunctionType::get(unwrap(builder)->getContext(), {}, {}), mlir::ArrayAttr()
+  );
+  contract.getBody().emplaceBlock();
+  return contract;
+}
+
+} // namespace
+
+TEST_F(CAPITest, llzk_verif_include_op_build_smoke) {
```
0xddom: Not the biggest of deal but the snake_case names are a hold over from when I ported the rust tests we had in `llzk-rs` to C++. Here they should be in camelCase

unittests/IR/VerifDialectTests.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+  OpBuilder builder(&ctx);
+  builder.setInsertionPointToStart(mod->getBody());
+
+  OperationState state(loc, ContractOp::getOperationName());
+
+  ContractOp::build(builder, state, "BuiltContract", "TargetFn");
+
+  auto *raw = Operation::create(state);
+  auto contract = cast<ContractOp>(raw);
```
0xddom: Why to use `builder.create`?

unittests/IR/VerifDialectTests.cpp
```diff
@@ -0,0 +1,256 @@
...
+  ASSERT_TRUE(succeeded(targetConstrain));
+
+  OpBuilder builder(&ctx);
+  builder.setInsertionPointToStart(mod->getBody());
+
+  OperationState state(loc, ContractOp::getOperationName());
+  ContractOp::build(builder, state, "StructContract", std::string(structNameA));
+
+  auto *raw = Operation::create(state);
+  auto contract = cast<ContractOp>(raw);
```
0xddom: Here as well. The whole point of the builder is not having to manually do this whole dance each time an op is created

include/llzk/Dialect/Verif/IR/Ops.td
```diff
@@ -0,0 +1,390 @@
...
+
+    Contract argument rules:
+    - For function targets, the contract accepts arguments matching the function's arguments.
+    - For struct targets, the contract accepts arguments matching the argument list of the
+    struct's `@constrain` function.
+
+    Contracts do not return values. For function targets with return values, the
+    return values are appended to the end of the contract's argument list.
+
+    Contracts may not directly target struct functions; they must target the structs themselves.
```
0xddom, line 104: Is this something we check during validation?

lib/Dialect/Verif/IR/Ops.cpp
0xddom: Right, I see that these methods are part of an interface. I'd raise a warning if this function returns a null pointer due to the op being managed, with a recommendation of inlining includes that may be causing it (since that's the only way that could happen)

0xddom (changes requested):
Only thing I'm not convinced is on the validations performed inside the build method. Other than that lgtm

lib/Dialect/Verif/IR/Ops.cpp
iangneal: So the goal of this builder was to simplify the interface by having the builder do the target lookup internally. If we don't want to do any validation in this builder (which we need to do in order to get the target and fetch the argument types), then we should just remove this builder. Alternatively, we could remove the error reporting and have the later verification methods trip the actual errors.

lib/Dialect/Verif/IR/Ops.cpp
iangneal: This is a hold-over from this being copied from the MLIR lib. I'll simplify.

lib/Dialect/Verif/IR/Ops.cpp
iangneal:
```
if (!callee->get().hasStructTarget()) {
    return nullptr;
  }
  ```
Already returns `nullptr` if the target is not a struct (i.e., is a function). So this assert is only hit if the target is a struct, and ergo must have a self argument.

include/llzk/Dialect/Verif/IR/Ops.td
iangneal: Yup. It's in the `isValidTarget` function in Ops.cpp. I'll add some explicit tests for this though, I think I neglected to add those.

lib/Dialect/Verif/IR/Ops.cpp
0xddom: I'm in favor of the latter. Convenience builders are always good when an operation is hard to construct. And, while I'm not sure if it is part of MLIR's core philosophy, I think one should be allowed to construct invalid IR for situations where the IR has incomplete information at the time of creation then gets fixed in a later step and have the validator scream at me if something was wrong. Conversion passes work like that; the IR is only validated after the pass finishes.

lib/Dialect/Verif/IR/Ops.cpp
0xddom: You are correct. I missed the forest while looking at the tree

0xddom (commented):
Last changes LGTM. Only thing left is that build method with validations in it

0xddom:
Question: Are ops for defining invariants going to be added in this PR or in a separate one? Are they going to be ops at all?

iangneal:
> Question: Are ops for defining invariants going to be added in this PR or in a separate one? Are they going to be ops at all?

We can add them in a subsequent PR if we need them. I was actually thinking that `invariant` can be lowered to a `bool.assert` inside of a loop or something similar, though if that gets annoying in the frontend, we can just add an op and do the lowering internally.

iangneal:
@0xddom resolved the last comment (removed verification in the custom builder, added a unit test to show that verification will then fail for an invocation of the custom builder that uses a bad target)

0xddom (approved):
LGTM!

PR #481, opened 2026-05-23
Add FeltConstantOp value access helpers

1sgtpepper:
## Summary

Closes #346.

Adds convenience accessors for `FeltConstantOp` so callers do not need to unwrap the stored `FeltConstAttr` manually.

This patch adds:
- `getValueAPInt()` for the underlying APInt value
- `tryGetZExtValue()` for checked 64-bit access
- one internal use of `getValueAPInt()` in felt constant result naming

The helpers use trailing return types so llzk-tblgen does not generate lossy C API wrappers for the APInt-returning method.

## Testing

- `git diff --check`
- Verified generated `Ops.capi.*.inc` does not expose `FeltConstantOp` wrappers for the new helpers

Full build/test coverage is left to CI.


iangneal (approved):
LGTM!

PR #486, opened 2026-05-27
Fix R1CS quadratic equality sign lowering

1sgtpepper:
## Summary

Fix R1CS lowering for quadratic-equals-linear constraints so the emitted `r1cs.constrain` preserves the linear side's sign.

## Details

`r1cs.constrain` enforces `a * b - c = 0`. The previous lowering path represented a quadratic side as the multiplicative operands, then subtracted the linear side into the internal `c` field. For a source equality like `constrain.eq (a * b), out`, that emitted `c = -out`, which enforced `a * b - (-out) = 0`.

This changes equality lowering so when either side is quadratic, the quadratic side stays as the multiplicative side and the opposite linear side is emitted with the sign needed by `a * b - c = 0`.

## Tests

- Updated existing R1CS lowering FileCheck expectations.
- Added a regression covering both operand orders:
  - `constrain.eq %prod_lhs, %lhs_out`
  - `constrain.eq %rhs_out, %prod_rhs`
- `git diff --check`


shankarapailoor (approved):
LGTM

PR #487, opened 2026-05-28
Bounds-check constant `array.len` dimensions

### LINKED ISSUE #482: array.len accepts out-of-range constant dimension and aborts in -llzk-array-to-scalar
## Summary

`array.len` accepts a statically out-of-range dimension index for a ranked array type. The accepted IR then aborts in `-llzk-array-to-scalar`, because the pass assumes the dimension index is valid and indexes the array type's dimension list without a bounds check.

For `!array.type<2,3 x !felt.type>`, valid dimension indices are `0` and `1`. The reproducer below asks for dimension `2`, which does not exist.

## Reproducer

```sh
cat > repro-array-len-oob.llzk <<'EOF_REPRO'
  function.def @bad_len() -> index {
    %a = array.new : !array.type<2,3 x !felt.type>
    %dim = arith.constant 2 : index
    %len = array.len %a, %dim : !array.type<2,3 x !felt.type>
    function.return %len : index
  }
}
EOF_REPRO

llzk-opt repro-array-len-oob.llzk
llzk-opt -llzk-array-to-scalar repro-array-len-oob.llzk
```

## Expected behavior

Either:

1. `array.len` should reject a statically known dimension index that is outside the array rank, or
2. `-llzk-array-to-scalar` should report a clean diagnostic instead of aborting.

For `!array.type<2,3 x !felt.type>`, dimension `0` has length `2` and dimension `1` has length `3`. Dimension `2` is out of range.

## Actual behavior

Plain `llzk-opt` accepts and prints the IR:

```text
  function.def @bad_len() -> index {
    %array = array.new  : <2,3 x !felt.type>
    %c2 = arith.constant 2 : index
    %0 = array.len %array, %c2 : <2,3 x !felt.type>
    function.return %0 : index
  }
}
```

Running `-llzk-array-to-scalar` aborts:

```text
llzk-opt: /usr/lib/llvm-20/include/llvm/ADT/ArrayRef.h:260:
Assertion `Index < Length && "Invalid index!"' failed.

#12 ReplaceKnownArrayLengthOp::getDimSizeIfKnown(...)
    lib/Dialect/Array/Transforms/ArrayToScalarPass.cpp:550:33
#13 ReplaceKnownArrayLengthOp::legal(...)
    lib/Dialect/Array/Transforms/ArrayToScalarPass.cpp:561:64
```

## Relevant code

`array.len` is documented as returning the size of the specified dimension:

https://github.com/project-llzk/llzk-lib/blob/1a62ceb98a0b2dd8af2b567b10ff7572096a6a20/include/llzk/Dialect/Array/IR/Ops.td#L379-L393

The documented example uses dimension `0` and dimension `1` for `!array.type<2,3 x !felt.type>`:

```mlir
%a = array.new : !array.type<2,3 x !felt.type>
%0 = arith.constant 0 : index
%x = array.len %a, %0 : !array.type<2,3 x !felt.type> // result is 2
%1 = arith.constant 1 : index
%y = array.len %a, %1 : !array.type<2,3 x !felt.type> // result is 3
```

`ArrayLengthOp` verification currently only checks symbol/type resolution:

https://github.com/project-llzk/llzk-lib/blob/1a62ceb98a0b2dd8af2b567b10ff7572096a6a20/lib/Dialect/Array/IR/Ops.cpp#L470-L472

The array-to-scalar pass then assumes the constant dimension is in range:

https://github.com/project-llzk/llzk-lib/blob/1a62ceb98a0b2dd8af2b567b10ff7572096a6a20/lib/Dialect/Array/Transforms/ArrayToScalarPass.cpp#L544-L561

In particular:

```cpp
size_t idx = llzk::checkedCast<size_t>(idxAP.getZExtValue());
Attribute dimSizeAttr = baseArrType.getDimensionSizes()[idx];
```

That is the aborting access when `idx == 2` for a rank-2 array.

A separate interpreter path treats this condition as invalid and reports `array.len dimension out of bounds`:

https://github.com/project-llzk/llzk-lib/blob/1a62ceb98a0b2dd8af2b567b10ff7572096a6a20/tools/llzk-witgen/Interpreter.cpp#L754-L770

## Impact

This lets verifier-accepted IR violate a basic array invariant: the selected dimension must exist. A downstream pass then trusts that invariant and aborts. This affects compiler robustness and makes `-llzk-array-to-scalar` unsafe to run on IR that the verifier currently accepts.

## Suggested fix direction

Add verifier coverage for constant dimension operands in `ArrayLengthOp`:

- if the dimension operand is a constant integer, reject values outside `[0, rank)`;
- keep dynamic dimension operands legal if they are intentionally supported;
- harden `ReplaceKnownArrayLengthOp::getDimSizeIfKnown()` with a bounds check before indexing `getDimensionSizes()`.

A regression test could be added to `test/Dialect/Array/array_len_fail.llzk` for the constant out-of-range case, plus a pass test to ensure `-llzk-array-to-scalar` no longer aborts.

## Additional verification detail

The in-range cases match the documented example: for `!array.type<2,3 x !felt.type>`, dimension `0` corresponds to length `2`, and dimension `1` corresponds to length `3`. The crash is specific to the statically known out-of-range constant dimension `2`.

Dynamic dimension operands may still be intentionally valid IR. The verifier-side fix can be limited to constant dimension operands where the value is known and outside `[0, rank)`, while the lowering pass should still defensively avoid indexing `getDimensionSizes()` without checking the rank first.

1sgtpepper:
Fixes #482.

## Summary
- Reject statically known `array.len` dimension operands outside `[0, rank)`
- Preserve support for dynamic dimension operands
- Add a defensive bounds check in `-llzk-array-to-scalar` before reading array dimension sizes
- Add verifier regressions for out-of-range and negative constant dimensions
- Add a no-crash array-to-scalar regression for malformed IR when verification is bypassed

## Testing
- Rebuilt `llzk-opt`
- Ran `test/Dialect/Array/array_len_fail.llzk`
- Ran `test/Dialect/Array/array_len_pass.llzk`
- Ran `test/Transforms/ArrayToScalar/array_to_scalar.llzk`
- Ran `test/Transforms/ArrayToScalar/array_len_invalid_dim_no_crash.llzk`
- Ran `git diff --check`

### DIFF
--- a/lib/Dialect/Array/IR/Ops.cpp
+++ b/lib/Dialect/Array/IR/Ops.cpp
@@ -20,6 +20,7 @@
 #include <mlir/IR/Attributes.h>
 #include <mlir/IR/BuiltinOps.h>
 #include <mlir/IR/Diagnostics.h>
+#include <mlir/IR/Matchers.h>
 #include <mlir/IR/OwningOpRef.h>
 #include <mlir/IR/SymbolTable.h>
 #include <mlir/IR/ValueRange.h>
@@ -27,6 +28,8 @@
 
 #include <llvm/ADT/ArrayRef.h>
 #include <llvm/ADT/Twine.h>
+
+#include <optional>
 
 // TableGen'd implementation files
 #include "llzk/Dialect/Array/IR/OpInterfaces.cpp.inc"
@@ -469,7 +472,27 @@ LogicalResult InsertArrayOp::verify() {
 
 LogicalResult ArrayLengthOp::verifySymbolUses(SymbolTableCollection &tables) {
   // Ensure any SymbolRef used in the type are valid
-  return verifyTypeResolution(tables, *this, getArrRefType());
+  if (failed(verifyTypeResolution(tables, *this, getArrRefType()))) {
+    return failure();
+  }
+
+  llvm::APInt dim;
+  if (!matchPattern(getDim(), m_ConstantInt(&dim))) {
+    return success();
+  }
+
+  std::optional<int64_t> idx = dim.trySExtValue();
+  if (!idx || *idx < 0 || static_cast<size_t>(*idx) >= getArrRefType().getDimensionSizes().size()) {
+    InFlightDiagnostic diag = emitOpError("dimension index ");
+    if (idx) {
+      diag << *idx;
+    } else {
+      diag << "outside the supported index range";
+    }
+    return diag << " is out of bounds for array rank "
+                << getArrRefType().getDimensionSizes().size();
+  }
+  return success();
 }
 
 } // namespace llzk::array
--- a/lib/Dialect/Array/IR/Ops.cpp
+++ b/lib/Dialect/Array/IR/Ops.cpp
@@ -482,15 +482,16 @@ LogicalResult ArrayLengthOp::verifySymbolUses(SymbolTableCollection &tables) {
   }
 
   std::optional<int64_t> idx = dim.trySExtValue();
-  if (!idx || *idx < 0 || static_cast<size_t>(*idx) >= getArrRefType().getDimensionSizes().size()) {
+  size_t rank = getArrRefType().getDimensionSizes().size();
+  if (!idx || *idx < 0 || llzk::checkedCast<size_t>(*idx) >= rank) {
     InFlightDiagnostic diag = emitOpError("dimension index ");
     if (idx) {
       diag << *idx;
     } else {
       diag << "outside the supported index range";
     }
-    return diag << " is out of bounds for array rank "
-                << getArrRefType().getDimensionSizes().size();
+    diag << " is out of bounds for array rank " << rank;
+    return diag.attachNote(getArrRef().getLoc()).append("array defined here");
   }
   return success();
 }
--- a/test/Transforms/ArrayToScalar/array_len_invalid_dim_no_crash.llzk
+++ b/test/Transforms/ArrayToScalar/array_len_invalid_dim_no_crash.llzk
@@ -1,14 +1,14 @@
-// RUN: llzk-opt --mlir-very-unsafe-disable-verifier-on-parsing --verify-each=false -llzk-array-to-scalar %s | FileCheck %s
+// RUN: llzk-opt -llzk-array-to-scalar %s -verify-diagnostics
 
-// This intentionally malformed IR checks that array-to-scalar defensively avoids
-// indexing past the array rank if verification is bypassed.
+// This intentionally malformed IR checks that array-to-scalar reports the
+// expected verifier diagnostic for an invalid constant dimension.
 module attributes {llzk.lang} {
   function.def @bad_len() -> index {
+    // expected-note@+1 {{array defined here}}
     %a = array.new : !array.type<2,3 x !felt.type>
     %dim = arith.constant 2 : index
+    // expected-error@+1 {{dimension index 2 is out of bounds for array rank 2}}
     %len = array.len %a, %dim : !array.type<2,3 x !felt.type>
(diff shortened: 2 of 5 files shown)


lib/Dialect/Array/IR/Ops.cpp
```diff
@@ -469,7 +472,27 @@ LogicalResult InsertArrayOp::verify() {
...
+    return failure();
+  }
+
+  llvm::APInt dim;
+  if (!matchPattern(getDim(), m_ConstantInt(&dim))) {
+    return success();
+  }
+
+  std::optional<int64_t> idx = dim.trySExtValue();
+  if (!idx || *idx < 0 || static_cast<size_t>(*idx) >= getArrRefType().getDimensionSizes().size()) {
```
iangneal: For consistency.
```suggestion
  if (!idx || *idx < 0 || llzk::checkedCast<size_t>(*idx) >= getArrRefType().getDimensionSizes().size()) {
```

lib/Dialect/Array/IR/Ops.cpp
```diff
@@ -469,7 +472,27 @@ LogicalResult InsertArrayOp::verify() {
...
+    InFlightDiagnostic diag = emitOpError("dimension index ");
+    if (idx) {
+      diag << *idx;
+    } else {
+      diag << "outside the supported index range";
+    }
+    return diag << " is out of bounds for array rank "
+                << getArrRefType().getDimensionSizes().size();
```
iangneal:
```suggestion
    diag << " is out of bounds for array of rank "
                << getArrRefType().getDimensionSizes().size();
    return diag.attachNote(getArrRef()->getLoc()).append("array defined here");
```
Just for added diagnostics. You may need to tweak that syntax.

test/Transforms/ArrayToScalar/array_len_invalid_dim_no_crash.llzk
```diff
@@ -0,0 +1,14 @@
...
+module attributes {llzk.lang} {
+  function.def @bad_len() -> index {
+    %a = array.new : !array.type<2,3 x !felt.type>
+    %dim = arith.constant 2 : index
+    %len = array.len %a, %dim : !array.type<2,3 x !felt.type>
+    function.return %len : index
+  }
+}
+
+// CHECK: array.len
```
iangneal: Checking this output is not very meaningful on its own. We should just go with a `-verify-diangostics` check so that we confirm we see the expected error.

```suggestion
// RUN: llzk-opt -llzk-array-to-scalar %s -verify-diagnostics

// This intentionally malformed IR checks that array-to-scalar defensively avoids
// indexing past the array rank if verification is bypassed.
module attributes {llzk.lang} {
  function.def @bad_len() -> index {
    %a = array.new : !array.type<2,3 x !felt.type>
    %dim = arith.constant 2 : index
    // expected-error@+1{{'array.len' op dimension index 2 is out of bounds for array rank 2}}
    %len = array.len %a, %dim : !array.type<2,3 x !felt.type>
    function.return %len : index
  }
}
```

iangneal (changes requested):
Hey @Kuhai9801, thanks for your PR! I've added a few suggestions.

1sgtpepper:
Thanks @iangneal, I addressed the review comments.

Updated the bounds check to use `llzk::checkedCast<size_t>` for consistency, added the attached diagnostic note pointing to the array definition, and changed the transform regression to use `-verify-diagnostics` with the expected error and note.

I kept the diagnostic wording as “array rank” to match the existing verifier tests.

iangneal (approved):
LGTM!

PR #489, opened 2026-05-28
Keep nondet ops distinct in duplicate elimination

### LINKED ISSUE #488: `llzk-duplicate-op-elim` merges distinct `llzk.nondet` witnesses
## Summary

`--llzk-duplicate-op-elim` treats distinct `llzk.nondet` operations as CSE-equivalent and rewrites later uses to the first nondet result. This breaks the documented `llzk.nondet` semantics and can change the constraint relation by forcing independent witness variables to be equal.

## Root cause

The pass keys duplicate operations with `OperationComparator`, then replaces a dominated matching operation's results with the earlier result. There is no positive CSE eligibility gate for freshness, effects, determinism, or result compatibility.

That conflicts with the `llzk.nondet` contract: `LLZK_NonDetOp` explicitly avoids `ConstantLike`/`NoMemoryEffect` because distinct nondet SSA values may be constrained differently.

## Reproduction

```mlir
  struct.def @NondetCSE {
    function.def @compute(%a: !felt.type, %b: !felt.type) -> !struct.type<@NondetCSE> {
      %self = struct.new : !struct.type<@NondetCSE>
      function.return %self : !struct.type<@NondetCSE>
    }
    function.def @constrain(%self: !struct.type<@NondetCSE>, %a: !felt.type, %b: !felt.type) {
      %x = llzk.nondet : !felt.type
      %y = llzk.nondet : !felt.type
      constrain.eq %x, %a : !felt.type
      constrain.eq %y, %b : !felt.type
      function.return
    }
  }
}
```

Run:

```bash
llzk-opt -split-input-file -llzk-duplicate-op-elim nondet_cse.llzk
```

Observed output from `llzk-opt` 2.0.0:

```mlir
%nondet = llzk.nondet : !felt.type
constrain.eq %nondet, %arg1 : !felt.type, !felt.type
constrain.eq %nondet, %arg2 : !felt.type, !felt.type
```

The two source nondets became one SSA value.

## Expected behavior

The pass should leave both nondet operations distinct. CSE should only apply to operations that are proven deterministic, non-fresh, effect-free, and value-interchangeable.

## Impact

The source relation has two independent existential witnesses. The optimized relation requires one witness to satisfy both constraints. For example, if `%a != %b`, the optimized constraints are no longer equivalent to the source constraints.

This is a semantic-preservation issue in an optimizer pass used before downstream lowering.

## Fix direction

Add a conservative CSE eligibility predicate before inserting/matching operations in `uniqueOps`. At minimum, exclude `llzk.nondet` and other freshness-producing operations, require a positive effect/determinism proof, and verify result arity/types before remapping results.

## Regression test

Add a duplicate-op-elim test asserting two `llzk.nondet` operations remain after the pass:

```mlir
// CHECK-LABEL: function.def @constrain
// CHECK: llzk.nondet
// CHECK: llzk.nondet
```

1sgtpepper:
Fixes #488.

## Summary
- Skip duplicate-op elimination for `llzk.nondet`
- Add a regression proving distinct nondet witnesses remain distinct
- Add an unreleased changelog entry

## Testing
- `clang-format`
- `git diff --check`

### DIFF
--- a/changelogs/unreleased/fix-duplicate-op-elim-nondet.yaml
+++ b/changelogs/unreleased/fix-duplicate-op-elim-nondet.yaml
@@ -1,0 +1,2 @@
+fixed:
+  - Fixed `llzk-duplicate-op-elim` so it does not merge distinct `llzk.nondet` witness values.
--- a/lib/Transforms/LLZKRedundantOperationEliminationPass.cpp
+++ b/lib/Transforms/LLZKRedundantOperationEliminationPass.cpp
@@ -15,6 +15,7 @@
 #include "llzk/Analysis/CallGraphAnalyses.h"
 #include "llzk/Dialect/Bool/IR/Ops.h"
 #include "llzk/Dialect/Constrain/IR/Ops.h"
+#include "llzk/Dialect/LLZK/IR/Ops.h"
 #include "llzk/Dialect/Struct/IR/Dialect.h"
 #include "llzk/Transforms/LLZKTransformationPasses.h"
 #include "llzk/Util/SymbolHelper.h"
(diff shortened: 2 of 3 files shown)


iangneal (approved):
LGTM!

PR #496, opened 2026-05-29
Reject unsupported zero-result ops in ZKLean conversion

### LINKED ISSUE #495: LLZK-to-ZKLean conversion silently drops zero-result constraint operations
### Summary

`--convert-llzk-to-zklean` can silently erase constraint operations that have no SSA results.

I found two cases:

1. `constrain.in` is dropped from the converted function body.
2. `function.call` to another `function.allow_constraint` function is dropped from the caller body.

In both cases the pass exits successfully and emits ZKLean IR, but the emitted artifact has fewer constraints than the source LLZK program.

### Reproduction

Tested against `project-llzk/llzk-lib` main at:

```text
30b0fa1eb77de154ff60c13fa88ef286d8b01c65
```

#### Case 1: `constrain.in` is dropped

Input:

```llzk
module attributes {llzk.lang = "llzk"} {
  function.def @contains(%xs: !array.type<2 x !felt.type<"babybear">>, %x: !felt.type<"babybear">) attributes {function.allow_constraint} {
    constrain.in %xs, %x : !array.type<2 x !felt.type<"babybear">>, !felt.type<"babybear">
    function.return
  }
}
```

Run:

```sh
llzk-opt repro-constrain-in.llzk --convert-llzk-to-zklean
```

Observed output:

```mlir
module @ZKLean attributes {llzk.lang = "llzk"} {
  func.func @contains(%arg0: !array.type<2 x !felt.type<"babybear">>, %arg1: !felt.type<"babybear">) {
    return
  }
}
```

The source function contains a `constrain.in`, but the converted ZKLean function body contains only `return`.

#### Case 2: constraint helper call is dropped from caller

Input:

```llzk
module attributes {llzk.lang = "llzk"} {
  struct.def @Child {
    struct.member @out : !felt.type<"babybear">

    function.def @compute(%x: !felt.type<"babybear">) -> !struct.type<@Child> {
      struct.writem %self[@out] = %x : !struct.type<@Child>, !felt.type<"babybear">
    }

    function.def @constrain(%self: !struct.type<@Child>, %x: !felt.type<"babybear">) attributes {function.allow_constraint} {
      %one = felt.const 1 <"babybear">
      constrain.eq %x, %one : !felt.type<"babybear">
      function.return
    }
  }

  function.def @parent(%child: !struct.type<@Child>, %x: !felt.type<"babybear">) attributes {function.allow_constraint} {
    function.call @Child::@constrain(%child, %x) : (!struct.type<@Child>, !felt.type<"babybear">) -> ()
    function.return
  }
}
```

Run:

```sh
llzk-opt repro-call.llzk --convert-llzk-to-zklean
```

Observed output includes a converted child constraint:

```mlir
func.func @Child__constrain(%arg0: !ZKLeanLean.type<@Child>, %arg1: !felt.type<"babybear">) {
  %felt_const_1 = felt.const  1 : <"babybear">
  %0 = ZKExpr.Literal %felt_const_1
  %1 = ZKExpr.Literal %arg1
  %2 = ZKBuilder.ConstrainEq %1 %0
  return
}
```

But the caller body is empty:

```mlir
func.func @parent(%arg0: !ZKLeanLean.type<@Child>, %arg1: !felt.type<"babybear">) {
  return
}
```

So `@Child__constrain` is emitted, but the `@parent` function no longer enforces the call to it.

### Expected behavior

The converter should preserve constraint semantics, or fail loudly when it cannot.

For unsupported zero-result operations, I would expect something like:

- emit a diagnostic,
- set the conversion error state,
- fail `--convert-llzk-to-zklean`.

For supported operations, the ZKLean output should contain equivalent constraints. In particular, a `function.call` from one `function.allow_constraint` function to another should not disappear from the caller body.

### Actual behavior

The pass exits successfully and emits ZKLean IR.

Unsupported zero-result operations appear to fall through without producing output and without causing conversion failure. This is different from unsupported result-producing operations, which do report an unsupported-op error.

### Impact

The converted ZKLean artifact can be weaker than the LLZK source program.

This matters because the output of `--convert-llzk-to-zklean` may look like a successful translation, while constraints from the source have been erased. A downstream consumer of the generated ZKLean IR would see a different constraint system from the one expressed in LLZK.

### Likely fix area

The issue appears to be in the per-operation conversion logic here:

```text
backends/zklean/lib/Conversions/LLZKToZKLean.cpp
```

The current conversion handles known operations such as `felt.const`, arithmetic ops, `struct.readm`, and `constrain.eq`. At the end, unsupported operations are only rejected when `op->getNumResults() != 0`.

That leaves unsupported zero-result operations, including `constrain.in` and `function.call`, able to fall through silently.

A minimal fix would be to explicitly reject unhandled zero-result operations, excluding only operations that are intentionally no-ops such as `function.return`. A fuller fix could lower `constrain.in` and constraint-function calls where those are intended to be supported.

1sgtpepper:
### Summary

Fixes #495.

`--convert-llzk-to-zklean` previously rejected unsupported operations only when they produced SSA results. Unsupported zero-result operations could therefore fall through without diagnostics, allowing constraint operations such as `constrain.in` or zero-result `function.call` to be omitted from the emitted ZKLean module while conversion still succeeded.

This change makes the converter fail on every unhandled operation except an empty `function.return`.

### Changes

- Treat only zero-operand `function.return` as an intentional no-op during LLZK-to-ZKLean conversion.
- Emit an unsupported-op diagnostic for all other unhandled operations, including zero-result ops.
- Reject multi-block LLZK functions before conversion, since the converter only processes a single block.
- Add regression coverage for:
  - `constrain.in`
  - zero-result constraint helper calls
  - result-bearing `function.return`

### Testing

- `git diff --check`

### DIFF
--- a/backends/zklean/lib/Conversions/LLZKToZKLean.cpp
+++ b/backends/zklean/lib/Conversions/LLZKToZKLean.cpp
@@ -421,9 +421,10 @@ static bool convertFunction(llzk::function::FuncDefOp func, LLZKToZKLeanState &s
 
 // Convert all eligible LLZK functions into a new ZKLean module.
 // Emits struct defs first, then lowers each function via `convertFunction`.
-static LogicalResult convertModule(ModuleOp source, ModuleOp dest, bool &hadError) {
+static LogicalResult convertModule(ModuleOp source, ModuleOp dest) {
   OpBuilder builder(dest.getContext());
   auto zkType = llzk::zkexpr::ZKExprType::get(dest.getContext());
+  bool hadError = false;
   bool createdAny = false;
 
   LLZKToZKLeanState state {dest, builder, zkType, hadError};
@@ -436,11 +437,11 @@ static LogicalResult convertModule(ModuleOp source, ModuleOp dest, bool &hadErro
   });
 
   if (!hadError && !createdAny) {
-    source.emitError("failed to produce ZKLean module").report();
+    source.emitError("did not produce any ops in ZKLean module").report();
     hadError = true;
   }
 
-  return success(!hadError && createdAny);
+  return failure(hadError);
 }
 
 // Pass wrapper that appends a converted ZKLean module to the source.
@@ -461,8 +462,7 @@ class PassImpl : public zklean::impl::ConvertLLZKToZKLeanPassBase<PassImpl> {
       zkLeanModule->setAttr(llzk::LANG_ATTR_NAME, lang);
     }
 
-    bool hadError = false;
-    if (failed(convertModule(original, zkLeanModule, hadError))) {
+    if (failed(convertModule(original, zkLeanModule))) {
       signalPassFailure();
       return;
     }
--- a/backends/zklean/lib/Conversions/LLZKToZKLean.cpp
+++ b/backends/zklean/lib/Conversions/LLZKToZKLean.cpp
@@ -421,11 +421,10 @@ static bool convertFunction(llzk::function::FuncDefOp func, LLZKToZKLeanState &s
 
 // Convert all eligible LLZK functions into a new ZKLean module.
 // Emits struct defs first, then lowers each function via `convertFunction`.
-static LogicalResult convertModule(ModuleOp source, ModuleOp dest) {
+static LogicalResult convertModule(ModuleOp source, ModuleOp dest, bool &hadError) {
   OpBuilder builder(dest.getContext());
   auto zkType = llzk::zkexpr::ZKExprType::get(dest.getContext());
   bool createdAny = false;
-  bool hadError = false;
 
   LLZKToZKLeanState state {dest, builder, zkType, hadError};
   emitZKLeanStructDefs(source, state);
@@ -435,6 +434,11 @@ static LogicalResult convertModule(ModuleOp source, ModuleOp dest) {
       createdAny = true;
     }
   });
+
+  if (!hadError && !createdAny) {
+    source.emitError("failed to produce ZKLean module").report();
+    hadError = true;
+  }
 
   return success(!hadError && createdAny);
 }
@@ -457,8 +461,8 @@ class PassImpl : public zklean::impl::ConvertLLZKToZKLeanPassBase<PassImpl> {
       zkLeanModule->setAttr(llzk::LANG_ATTR_NAME, lang);
     }
 
-    if (failed(convertModule(original, zkLeanModule))) {
-      original.emitError("failed to produce ZKLean module").report();
+    bool hadError = false;
+    if (failed(convertModule(original, zkLeanModule, hadError))) {
       signalPassFailure();
       return;
     }
--- a/backends/zklean/lib/Conversions/LLZKToZKLean.cpp
+++ b/backends/zklean/lib/Conversions/LLZKToZKLean.cpp
@@ -377,10 +377,13 @@ struct FunctionConverter {
       state.builder.create<llzk::zkbuilder::ConstrainEqOp>(eq.getLoc(), stateType, lhs, rhs);
       return;
     }
-    if (op->getNumResults() != 0) {
-      op->emitError("unsupported op in ZKLean conversion").report();
-      state.hadError = true;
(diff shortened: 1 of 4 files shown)


backends/zklean/lib/Conversions/LLZKToZKLean.cpp
```diff
@@ -410,11 +418,10 @@ static bool convertFunction(llzk::function::FuncDefOp func, LLZKToZKLeanState &s
 
 // Convert all eligible LLZK functions into a new ZKLean module.
 // Emits struct defs first, then lowers each function via `convertFunction`.
-static LogicalResult convertModule(ModuleOp source, ModuleOp dest) {
+static LogicalResult convertModule(ModuleOp source, ModuleOp dest, bool &hadError) {
```
tim-hoffman: Doesn't seem like there's any reason to pass in `hadError`

backends/zklean/lib/Conversions/LLZKToZKLean.cpp
```diff
@@ -425,6 +432,11 @@ static LogicalResult convertModule(ModuleOp source, ModuleOp dest) {
     }
   });
 
+  if (!hadError && !createdAny) {
+    source.emitError("failed to produce ZKLean module").report();
+    hadError = true;
+  }
+
   return success(!hadError && createdAny);
```
tim-hoffman:
```suggestion
  if (!hadError && !createdAny) {
    source.emitError("did not produce any ops in ZKLean module").report();
    hadError = true;
  }

return failure(hadError);
```

tim-hoffman (dismissed):
Initial feedback

tim-hoffman:
@benoitrazet, this PR revealed a few scenarios that are not handled in the `llzk-to-zklean` conversion and I'm hoping you can give some feedback about what's expected from the zklean backend.

1. The transformation does not currently support structs that use other structs for sub-circuits (see: [circomlib.llzk](https://github.com/project-llzk/llzk-lib/blob/main/test/FrontendLang/Circom/circomlib.llzk) where the error  "unsupported member type for ZKLean struct conversion" occurs because of the `struct.type` member, and `convertOperation()` has no handling for `CallOp`). Should the transformation handle subcomponents directly or is it expected to run the `inline-structs` pass first? If the latter, I think we should just make it a pipeline that always runs that pass.
2. Since there is no handling for `CallOp`, free (non-struct) functions called from a `@constrain` function would also be dropped (see: [poseidon](https://github.com/project-llzk/llzk-lib/blob/8a76d0add470349efcccd29b0e48b11d657159eb/test/FrontendLang/Circom/poseidon_m_concrete.llzk#L75)). The `inline-structs` pass will not help in this case so the common question is, can the ZKLean output contain function calls? If not, we will need a pass to inline these as well (fitting into my idea of making it a pipeline if necessary).
3. The `constrain.in` op is not currently handled. Does ZKLean have a means to represent that op?

backends/zklean/lib/Conversions/LLZKToZKLean.cpp
```diff
@@ -395,6 +398,11 @@ static bool convertFunction(llzk::function::FuncDefOp func, LLZKToZKLeanState &s
   if (state.hadError || !shouldConvertFunc(func)) {
     return false;
   }
+  if (!func.getBody().hasOneBlock()) {
+    func.emitError("ZKLean conversion only supports single-block functions").report();
+    state.hadError = true;
+    return false;
+  }
```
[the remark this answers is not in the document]
1sgtpepper: Added the requested regression in `test/Conversions/llzk_to_zklean_unsupported_zero_result_fail.llzk`: the `@multi_block` case expects the single-block diagnostic. The rebased head includes it.

1sgtpepper:
I suggest keeping #496 focused on rejecting unsupported zero-result ops. Callers that need calls or subcomponents can run `llzk-full-inlining` before conversion; remaining unsupported calls will still fail. `constrain.in` needs a separate representation/lowering follow-up. The branch is now rebased onto current `main`.

shankarapailoor:
@codex review please

shankarapailoor (approved):
LGTM

PR #497, opened 2026-05-31
Validate function.res_name attributes

### LINKED ISSUE #490: Add checks for `function.res_name` attributes
Add checks and support for a `function.res_name` attribute so that function return results can be referred to by name.
- The attribute should be restricted to function return values
- It should only be a `StringAttr`

1sgtpepper:
## Summary

Add verifier support for `function.res_name` result attributes.

This mirrors the existing `function.arg_name` validation rules for function results:
- `function.res_name` is only valid on function results
- values must be non-empty, untyped `StringAttr`s
- result names must be unique within the function

Fixes #490.

## Testing

- `git diff --check`

### DIFF
--- a/changelogs/unreleased/function-res-name-attrs.yaml
+++ b/changelogs/unreleased/function-res-name-attrs.yaml
@@ -1,0 +1,2 @@
+added:
+  - Add verifier support for `function.res_name` result attributes.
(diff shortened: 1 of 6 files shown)


iangneal (approved):
LGTM! Thank you!

PR #500, opened 2026-06-01
Document felt shift operator semantics

1sgtpepper:
## Summary
- Document `felt.shl` and `felt.shr` semantics in the Felt dialect TableGen docs.
- Clarify that shifts use unsigned integer representatives and how results are converted back to field elements.

Closes #393

## Checks
- `git diff --check`


iangneal (approved):
LGTM!

PR #501, opened 2026-06-01
Fix nested member aux rebuild

### LINKED ISSUE #491: Poly lowering rebuilds nested member-read auxiliary expressions from the wrong component
# Poly lowering rebuilds nested member-read auxiliary expressions from the wrong component

## Summary

When polynomial lowering introduces an auxiliary member for an expression involving a nested member read, `rebuildExprInCompute` rebuilds every `struct.readm` from the enclosing compute function's `%self`. It does not rebuild the original read's component operand. A constrain-side path like `parent.child.val` can therefore become `parent.val` in `compute()`.

## Root cause

[`rebuildExprInCompute`](https://github.com/project-llzk/llzk-lib/blob/22931df604ba40ecf50c6260c90365df16a9c222/lib/Transforms/LLZKLoweringUtils.cpp#L42-L47) handles `MemberReadOp` by creating a new read from `computeFunc.getSelfValueFromCompute()` and the same member name. It should recursively rebuild `readOp.getComponent()` instead.

The helper is used when poly lowering writes generated auxiliary members [here](https://github.com/project-llzk/llzk-lib/blob/22931df604ba40ecf50c6260c90365df16a9c222/lib/Transforms/LLZKPolyLoweringPass.cpp#L317-L328), and the R1CS lowering path uses the same helper for aux writes [here](https://github.com/project-llzk/llzk-lib/blob/22931df604ba40ecf50c6260c90365df16a9c222/backends/r1cs/lib/r1cs/Transforms/R1CSLoweringPass.cpp#L685-L696).

## Reproduction

Use a parent and child that both have a compatible `@val` member:

```mlir
  struct.def @Child {
    struct.member @val : !felt.type {llzk.pub}
    function.def @compute(%x: !felt.type) -> !struct.type<@Child> {
      struct.writem %self[@val] = %x : !struct.type<@Child>, !felt.type
    }
  }
  struct.def @Parent {
    struct.member @child : !struct.type<@Child>
    struct.member @val : !felt.type
    function.def @compute(%x: !felt.type) -> !struct.type<@Parent> {
      %self = struct.new : !struct.type<@Parent>
      function.return %self : !struct.type<@Parent>
    }
    function.def @constrain(%self: !struct.type<@Parent>, %x: !felt.type) {
      %child = struct.readm %self[@child] : !struct.type<@Parent>, !struct.type<@Child>
      %child_val = struct.readm %child[@val] : !struct.type<@Child>, !felt.type
      %sq = felt.mul %child_val, %child_val
      %quartic = felt.mul %sq, %sq
      %parent_val = struct.readm %self[@val] : !struct.type<@Parent>, !felt.type
      constrain.eq %parent_val, %quartic : !felt.type
      function.return
    }
  }
}
```

Run:

```bash
llzk-opt -I . -split-input-file -llzk-full-poly-lowering nested_member_read_aux_rebuild.llzk
```

Observed output from `llzk-opt` 2.0.0 at `22931df604ba40ecf50c6260c90365df16a9c222`:

```mlir
function.def @compute(%arg0: !felt.type) -> !struct.type<@Parent> {
  %self = struct.new : <@Parent>
  %0 = struct.readm %self[@val] : <@Parent>, !felt.type
  %1 = felt.mul %0, %0 : !felt.type, !felt.type
  struct.writem %self[@__llzk_poly_lowering_pass_aux_member_0] = %1 : <@Parent>, !felt.type
  function.return %self : !struct.type<@Parent>
}
```

The lowered `constrain()` still reads `@child` and then child `@val`; only the generated compute-side aux assignment uses direct parent `@val`.

## Expected behavior

The compute-side auxiliary expression should preserve the member path shape: read `%self[@child]`, then read `@val` from that child, then compute the auxiliary value.

## Impact

The generated witness computation can disagree with the lowered constraints. In this example, the aux value is computed from `Parent.val^2`, while the constrain-side relation expects it to correspond to `Child.val^2`. This changes the semantics of lowered LLZK programs that use nested component fields in high-degree expressions.

## Fix direction

In `rebuildExprInCompute`, map the original constrain `%self` to compute `%self`, then recursively rebuild the original `MemberReadOp` component before recreating the read. Do not replace every member-read base with the enclosing compute `%self`.

## Regression test

Add a poly-lowering test with both `Parent.@val` and `Child.@val`. FileCheck should require a compute-side `struct.readm %self[@child]` followed by `struct.readm <child>[@val]`, and reject computing the aux from direct `Parent.@val`.

1sgtpepper:
Fixes #491.

## Summary

Fix compute-side auxiliary reconstruction for nested and call-rooted member reads during poly lowering.

This now:
- maps constrain `%self` to compute `%self`
- recursively rebuilds member-read components instead of rebuilding every read from compute `%self`
- propagates unsupported-root rebuild failures cleanly after diagnostics
- keeps R1CS explicit: self-rooted member reads are supported, non-self-rooted reads are rejected

## Testing

- `git diff --check upstream/main...HEAD`
- `git clang-format --diff upstream/main HEAD --binary clang-format`
- GitHub Actions on `ec90cfd`: all checks passing


test/Transforms/PolyLowering/poly_lowering_pass_deg2.llzk
```diff
@@ -193,11 +193,12 @@ module attributes {llzk.lang} {
...
-// CHECK:             %[[VAL_4:.*]] = felt.mul %[[VAL_3]], %[[VAL_3]] : !felt.type, !felt.type
-// CHECK:             struct.writem %[[VAL_2]][@__llzk_poly_lowering_pass_aux_member_1] = %[[VAL_4]] : <@Mod1>, !felt.type
-// CHECK:             %[[VAL_5:.*]] = felt.mul %[[VAL_0]], %[[VAL_1]] : !felt.type, !felt.type
-// CHECK:             struct.writem %[[VAL_2]][@__llzk_poly_lowering_pass_aux_member_2] = %[[VAL_5]] : <@Mod1>, !felt.type
+// CHECK:             %[[CHILD:.*]] = struct.readm %[[VAL_2]][@mod2] : <@Mod1>, !struct.type<@Mod2>
+// CHECK:             %[[CHILD_VAL:.*]] = struct.readm %[[CHILD]][@val] : <@Mod2>, !felt.type
+// CHECK:             %[[AUX_VAL:.*]] = felt.mul %[[CHILD_VAL]], %[[CHILD_VAL]] : !felt.type, !felt.type
+// CHECK:             struct.writem %[[VAL_2]][@__llzk_poly_lowering_pass_aux_member_1] = %[[AUX_VAL]] : <@Mod1>, !felt.type
+// CHECK:             %[[ARG_PRODUCT:.*]] = felt.mul %[[VAL_0]], %[[VAL_1]] : !felt.type, !felt.type
+// CHECK:             struct.writem %[[VAL_2]][@__llzk_poly_lowering_pass_aux_member_2] = %[[ARG_PRODUCT]] : <@Mod1>, !felt.type
```
tim-hoffman: @shankarapailoor can you verify this change is correct?

tim-hoffman:
@codex review

1sgtpepper:
I wanted to check the intended behavior for poly/R1CS lowering here. Should aux expressions that depend on nested-region block args, like `scf.for` IVs or `iter_args`, be supported?

`rebuildExprInCompute` currently maps any `BlockArgument` by ordinal to compute args, which seems correct for top-level `constrain` args but not for nested-region args. If that case is unsupported for now, I think it would be better to reject it explicitly with a diagnostic and add a regression test.

shankarapailoor:
@Kuhai9801 Since the poly lowering pass now has the explicit check that the program is straight-line, your above question is answered right?

1sgtpepper:
Thanks, that makes sense. Yes, #545 answers that part.

I’ll rebase this on current `main` and keep this PR focused on the straight-line nested member/call rebuild issue from #491. I’ll also mark it draft while I update the conflicts and trim anything now covered by the straight-line guard.

iangneal:
@codex review

backends/r1cs/lib/r1cs/Transforms/R1CSLoweringPass.cpp
```diff
@@ -479,22 +482,36 @@ class PassImpl : public r1cs::impl::R1CSLoweringPassBase<PassImpl> {
...
+                .report();
+            signalPassFailure();
+            return failure();
+          }
           auto memberVal = memberMap.find(read.getMemberName());
-          assert(memberVal != memberMap.end() && "Member read not associated with a value");
+          if (memberVal == memberMap.end()) {
+            read.emitError("member read is not associated with an R1CS signal").report();
+            signalPassFailure();
+            return failure();
```
tim-hoffman:
```suggestion
            signalPassFailure();
            return read.emitError("member read is not associated with an R1CS signal");
```

backends/r1cs/lib/r1cs/Transforms/R1CSLoweringPass.cpp
```diff
@@ -479,22 +482,36 @@ class PassImpl : public r1cs::impl::R1CSLoweringPassBase<PassImpl> {
...
         Operation *op = v.getDefiningOp();
         if (auto read = dyn_cast<MemberReadOp>(op)) {
+          if (read.getComponent() != selfVal) {
+            read.emitError(
+                    "R1CS lowering only supports member reads rooted at the current constrain "
+                    "self value"
+            )
+                .report();
+            signalPassFailure();
+            return failure();
```
tim-hoffman:
```suggestion
            signalPassFailure();
            return read.emitError(
                    "R1CS lowering only supports member reads rooted at the current constrain "
                    "self value"
            );
```

backends/r1cs/lib/r1cs/Transforms/R1CSLoweringPass.cpp
```diff
@@ -479,22 +482,36 @@ class PassImpl : public r1cs::impl::R1CSLoweringPassBase<PassImpl> {
...
+          if (memberVal == memberMap.end()) {
+            read.emitError("member read is not associated with an R1CS signal").report();
+            signalPassFailure();
+            return failure();
+          }
           return memberVal->second;
         }
         op->emitError("Value not mapped in R1CS lowering").report();
         signalPassFailure();
+        return failure();
```
tim-hoffman:
```suggestion
        signalPassFailure();
        return op->emitError("Value not mapped in R1CS lowering");
```

lib/Transforms/LLZKLoweringUtils.cpp
```diff
@@ -35,53 +37,126 @@ Value rebuildExprInCompute(
...
   }
 
   if (auto c = val.getDefiningOp<FeltConstantOp>()) {
     return memo[val] = builder.create<FeltConstantOp>(c.getLoc(), c.getValueAttr());
   }
 
-  llvm::errs() << "Unhandled op in rebuildExprInCompute: " << val << '\n';
-  llvm_unreachable("Unsupported op kind");
+  if (Operation *op = val.getDefiningOp()) {
+    op->emitError("cannot rebuild unsupported operation in compute-side auxiliary expression");
```
tim-hoffman:
```suggestion
    op->emitError("cannot rebuild unsupported operation in compute-side auxiliary expression").report();
```

lib/Transforms/LLZKLoweringUtils.cpp
```diff
@@ -35,53 +37,126 @@ Value rebuildExprInCompute(
...
+        readOp.getLoc(), readOp.getType(), component, readOp.getMemberNameAttr().getAttr()
     );
     return memo[val] = rebuilt;
   }
 
+  if (auto callOp = val.getDefiningOp<CallOp>()) {
+    if (!callOp.getMapOperands().empty()) {
+      callOp.emitError(
+          "cannot rebuild affine-instantiated function.call in compute-side auxiliary expression"
+      );
```
tim-hoffman: add defensive `report()` because in some cases the error will be dropped or misordered
```suggestion
      callOp.emitError(
          "cannot rebuild affine-instantiated function.call in compute-side auxiliary expression"
      ).report();
```

lib/Transforms/LLZKLoweringUtils.cpp
```diff
@@ -35,53 +37,126 @@ Value rebuildExprInCompute(
...
+    bool invalidTarget =
+        (targetFunc.hasAllowConstraintAttr() && !computeFunc.hasAllowConstraintAttr()) ||
+        (targetFunc.hasAllowWitnessAttr() && !computeFunc.hasAllowWitnessAttr()) ||
+        (targetFunc.hasAllowNonNativeFieldOpsAttr() &&
+         !computeFunc.hasAllowNonNativeFieldOpsAttr());
+    if (invalidTarget) {
+      callOp.emitError(
+          "cannot rebuild function.call in compute-side auxiliary expression because the callee "
+          "requires attributes not present on the compute function"
+      );
```
tim-hoffman:
```suggestion
      callOp.emitError(
          "cannot rebuild function.call in compute-side auxiliary expression because the callee "
          "requires attributes not present on the compute function"
      ).report();
```

test/Transforms/PolyLowering/poly_lowering_fail_unsupported_root.llzk
```diff
@@ -16,3 +16,23 @@ module attributes {llzk.lang} {
...
+module attributes {llzk.lang} {
+  struct.def @UnsupportedNondetAuxRoot {
+    function.def @compute(%a: !felt.type, %b: !felt.type, %out: !felt.type) -> !struct.type<@UnsupportedNondetAuxRoot> {
+      %self = struct.new : !struct.type<@UnsupportedNondetAuxRoot>
+      function.return %self : !struct.type<@UnsupportedNondetAuxRoot>
+    }
+
+    function.def @constrain(%self: !struct.type<@UnsupportedNondetAuxRoot>, %a: !felt.type, %b: !felt.type, %out: !felt.type) {
+      // expected-error@+1 {{cannot rebuild unsupported operation in compute-side auxiliary expression}}
+      %nondet = llzk.nondet : !felt.type
```
tim-hoffman: `llzk.nondet` can be used in "@compute" so it seems like this one need not produce an error

test/Transforms/PolyLowering/poly_lowering_pass_deg2.llzk
```diff
@@ -241,3 +242,51 @@ module attributes {llzk.lang, llzk.main = !struct.type<@Mod1>} {
...
+// CHECK-LABEL: module attributes {llzk.lang, llzk.main = !struct.type<@UsesExternPayload>} {
+// CHECK:        struct.def @UsesExternPayload {
+// CHECK:          function.def @compute(%[[ARG:[0-9a-zA-Z_\.]+]]: !felt.type) -> !struct.type<@UsesExternPayload> attributes {function.allow_witness} {
+// CHECK:            %[[SELF:[0-9a-zA-Z_\.]+]] = struct.new : <@UsesExternPayload>
+// CHECK:            %[[PAYLOAD:[0-9a-zA-Z_\.]+]] = function.call @extern_wrap(%[[ARG]]) : (!felt.type) -> !struct.type<@ExternPayload>
+// CHECK:            %[[VALUE:[0-9a-zA-Z_\.]+]] = struct.readm %[[PAYLOAD]][@value] : <@ExternPayload>, !felt.type
+// CHECK:            %[[SQUARE:[0-9a-zA-Z_\.]+]] = felt.mul %[[VALUE]], %[[VALUE]] : !felt.type, !felt.type
+// CHECK:            struct.writem %[[SELF]][@__llzk_poly_lowering_pass_aux_member_0] = %[[SQUARE]] : <@UsesExternPayload>, !felt.type
+// CHECK:            function.return %[[SELF]] : !struct.type<@UsesExternPayload>
+// CHECK:          }
```
tim-hoffman: prefer complete output check generated by the script

1sgtpepper:
@tim-hoffman aside from the requested fixes, I kept R1CS conservative for non-`%self` member reads and made redundant read/write elim tolerate unknown roots, so this PR only expands poly aux reconstruction without broadening R1CS or elim assumptions.

lib/Transforms/LLZKLoweringUtils.cpp
```diff
@@ -66,14 +68,76 @@ Value rebuildExprInCompute(
...
+        (targetFunc.hasAllowConstraintAttr() && !computeFunc.hasAllowConstraintAttr()) ||
+        (targetFunc.hasAllowWitnessAttr() && !computeFunc.hasAllowWitnessAttr()) ||
+        (targetFunc.hasAllowNonNativeFieldOpsAttr() &&
+         !computeFunc.hasAllowNonNativeFieldOpsAttr());
+    if (invalidTarget) {
+      callOp
+          .emitError(
+              "cannot rebuild function.call in compute-side auxiliary expression because the "
+              "callee "
+              "requires attributes not present on the compute function"
```
tim-hoffman: nit
```suggestion
              "callee requires attributes not present on the compute function"
```

test/Transforms/PolyLowering/poly_lowering_pass_deg2.llzk
```diff
@@ -154,7 +154,7 @@ module attributes {llzk.lang, llzk.main = !struct.type<@Mod1>} {
   // tests that polynomials are lowered to degree 1 before being passed into constraint calls
   struct.def @Mod1 {
     struct.member @mod2: !struct.type<@Mod2>
-    struct.member @val2: !felt.type
+    struct.member @val: !felt.type
```
tim-hoffman: please revert this `val2` to `val` rename

tim-hoffman (changes requested):

iangneal:
@codex review

tim-hoffman (approved):
LGTM. Thanks for your contribution!

PR #511, opened 2026-06-02
Lower composite poly roots

1sgtpepper:
## Summary

Fix poly lowering for high-degree felt expressions whose root op is linear.

`getDegree` already propagates degree through `felt.add`, `felt.sub`, and `felt.neg`, but `lowerExpression` only recursively lowered `felt.mul` roots. With `max-degree=2`, a constraint operand such as `(a*b*c)+d` could therefore keep a degree-3 add-root expression after lowering.

This change lowers the operands of add/sub/neg roots in place, so the root expression is preserved while high-degree children are materialized through the existing auxiliary-member path.

## Testing

- `git diff --check`
- Added `poly_lowering_composite_roots.llzk` covering add, sub, and neg roots with `llzk-poly-lowering-pass="max-degree=2"`

### DIFF
--- a/changelogs/unreleased/lower-composite-poly-roots.yaml
+++ b/changelogs/unreleased/lower-composite-poly-roots.yaml
@@ -1,0 +1,2 @@
+fixed:
+  - Lower high-degree composite felt expressions in the poly-lowering pass
--- a/changelogs/unreleased/lower-composite-poly-roots.yaml
+++ b/changelogs/unreleased/lower-composite-poly-roots.yaml
@@ -1,2 +1,2 @@
 fixed:
-  - Lower high-degree composite felt expressions in the poly-lowering pass
+  - Lower high-degree felt.add, felt.sub, and felt.neg equality roots and nonlinear struct constrain call arguments in the poly-lowering pass
--- a/include/llzk/Transforms/LLZKTransformationPasses.td
+++ b/include/llzk/Transforms/LLZKTransformationPasses.td
@@ -59,6 +59,10 @@ def PolyLoweringPass : LLZKPass<"llzk-poly-lowering-pass"> {
     Rewrites constraint expressions into an (observationally) equivalent system where the degree of
     every polynomial is less than or equal to the specified maximum.
 
+    High-degree subexpressions are factored into auxiliary struct members. The pass also recurses
+    through degree-neutral roots such as `felt.add`, `felt.sub`, and `felt.neg`, so composite
+    equality operands and struct constrain call arguments satisfy their degree bounds.
+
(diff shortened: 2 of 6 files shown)


shankarapailoor:
Lowering only multiplication ops was intentional because multiplication is the only op that can build expressions of higher degree. As such, could you explain how such an expression like a*b*c + d could be preserved by the pass since the multiplication op that produces a*b*c before the add should trigger the lowering. In particular, I would expect it to be lowered to an expression like y*c + d and an additional constraint of y = a*b to be created.

1sgtpepper:
Right, multiplication is the only degree-increasing op. The missing piece is that the old pass did not walk all nested `felt.mul` ops independently.

For equality constraints, `runOnOperation` calls `lowerExpression` on the top-level equality operand value:
https://github.com/project-llzk/llzk-lib/blob/6a73e3fbc991983721b77920cecc085478ed9fde/lib/Transforms/LLZKPolyLoweringPass.cpp#L256-L275

Before this PR, `lowerExpression` only recursed when that exact operand root was `MulFeltOp`. Non-mul roots fell through unchanged:
https://github.com/project-llzk/llzk-lib/blob/6a73e3fbc991983721b77920cecc085478ed9fde/lib/Transforms/LLZKPolyLoweringPass.cpp#L124-L131
https://github.com/project-llzk/llzk-lib/blob/6a73e3fbc991983721b77920cecc085478ed9fde/lib/Transforms/LLZKPolyLoweringPass.cpp#L209-L211

So for the reproducer, the equality RHS root is `%rhs = felt.add %abc, %d`, not `%abc`:
https://github.com/Kuhai9801/llzk-lib/blob/5df5c821b4a3991049d7feecbab096b469cc27a3/test/Transforms/PolyLowering/poly_lowering_composite_roots.llzk#L10-L14

That means the old pass computed the add-root expression as high-degree, called `lowerExpression(%rhs)`, then returned `%rhs` unchanged because the root was not `MulFeltOp`. The inner `%abc = felt.mul %ab, %c` was never reached through that path.

This PR keeps add/sub/neg as degree-neutral roots, but recursively lowers their operands, which gives the expected shape: materialize `%ab` as an aux value, then form `aux * c + d`.

I’m going to keep this as draft while I clean up the failing checks and strengthen the tests. In particular, I’ll add checks that the auxiliary assignments introduced for composite-root lowering are also rebuilt into `compute`, and I’ll add a struct-constrain call case where a composite-root felt expression is lowered and then materialized to a degree-1 call argument.

shankarapailoor (approved):
LGTM Thanks

PR #513, opened 2026-06-02
Update legalization for POD types and ops

### LINKED ISSUE #456: `-llzk-flatten` fails to legalize `pod.read`/`pod.write`
The flattening pass has not been updated since the inclusion of the POD type, so we just need to add support for it.

iangneal:
Fixes errors caused when POD ops/types appear in the empty template removal pass.


include/llzk/Dialect/Shared/TypeConversionPatterns.h
```diff
@@ -224,6 +227,33 @@ class CallOpClassReplacePattern : public mlir::OpConversionPattern<function::Cal
...
+      return op->emitError("Could not convert Op result types.");
+    }
+    mlir::OperandRangeRange orr = op.getMapOperands();
+    mlir::SmallVector<mlir::ValueRange> mapOperands;
+    mapOperands.reserve(orr.size());
+    mapOperands.insert(mapOperands.end(), orr.begin(), orr.end());
+    replaceOpWithNewOp<pod::NewPodOp>(
+        rewriter, op, newResultType, mapOperands, op.getNumDimsPerMapAttr(),
+        op.getInitializedRecordValues()
+    );
```
tim-hoffman: Anything containing SSA Values needs to come from the OpAdaptor rather than the original op so converted values are used rather than the original values; the map operands and initialized record values in this case.

tim-hoffman (changes requested):

tim-hoffman (commented):
Looks good but it would be nice to have a lit test where the `pod.new` op has an initial value parameter to cover the OpAdaptor usage case you just updated.

tim-hoffman (approved):
LGTM

PR #514, opened 2026-06-03
Add `llzk-pod-to-scalar` pass

### LINKED ISSUE #313: `InlineStructs` pass fails in the presence of POD read/write
When inlining a struct where the substruct result is written to a POD, during `combineNewThenReadChain`, the `findOpThatStoresSubcmp` fails because the result of the compute is written to a POD instead of directly to a struct member.

tim-hoffman:
fixes #313 ~and #456~

UPDATE: rather than supporting pod directly in `llzk-inline-structs`, the workflow should first run the `llzk-pod-to-scalar` pass added by this PR


include/llzk/Transforms/LLZKTransformationPasses.h
```diff
@@ -32,8 +32,6 @@ std::unique_ptr<mlir::Pass> createRedundantOperationEliminationPass();
 
 std::unique_ptr<mlir::Pass> createUnusedDeclarationEliminationPass();
 
-std::unique_ptr<mlir::Pass> createArrayToScalarPass();
-
```
tim-hoffman, line 36: removed because this was duplicated in the wrong header

iangneal:
@codex review

test/Transforms/InlineStructs/circom_subcmps5b.llzk
```diff
@@ -0,0 +1,130 @@
+// RUN: llzk-opt -llzk-flatten -llzk-pod-to-scalar -canonicalize -llzk-inline-structs %s 2>&1 | FileCheck --enable-var-scope %s
+//
+// COM: the following passes should run before inlining structs:
+// COM: - llzk-flatten: to instantiate templated structs and unroll loops
+// COM: - llzk-pod-to-scalar: to convert aggregate pod types into scalar SSA values
+// COM: - canonicalize: to remove known condition `scf.if` regions so struct inlining can link "@compute" calls to struct members
```
tim-hoffman, line 6: I believe canonicalize should be sufficient to get rid of `scf.if` around "@compute" calls for circom-generated code since the circom circuit structure must be static, not signal dependent, and IIRC the only non-source `scf.if` we produce are for early return from functions (N/A here) and the input signal initialization countdown (as shown in this test).

tim-hoffman:
@codex review

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1595 @@
...
+  return false;
+}
+
+/// If the given Type is a PodType that can be split into scalars, append `collect` with all of
+/// the scalar types that result from splitting the PodType. Otherwise, just push the `Type`.
+size_t splitPodTypeTo(Type t, SmallVector<Type> &collect) {
+  if (PodType pt = splittablePod(t)) {
+    auto records = pt.getRecords();
+    for (RecordAttr record : records) {
+      collect.push_back(record.getType());
```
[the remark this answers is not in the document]
tim-hoffman: Note: issue may be present in array-to-scalar pass also

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1595 @@
...
+static CallOp newCallOpWithSplitResults(
+    CallOp oldCall, CallOp::Adaptor adaptor, ConversionPatternRewriter &rewriter
+) {
+  OpBuilder::InsertionGuard guard(rewriter);
+  rewriter.setInsertionPointAfter(oldCall);
+
+  Operation::result_range oldResults = oldCall.getResults();
+  CallOp newCall = rewriter.create<CallOp>(
+      oldCall.getLoc(), splitPodType(oldResults.getTypes()), oldCall.getCallee(),
+      adaptor.getArgOperands()
```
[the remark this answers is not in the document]
tim-hoffman: Note: issue may be present in array-to-scalar pass also
Could give a warning stating the flattening pass must run first.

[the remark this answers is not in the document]
tim-hoffman: Note: issue may be present in array-to-scalar pass also
Could give a warning stating the flattening pass must run first.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1595 @@
...
+  } else {
+    return nullptr;
+  }
+}
+
+/// Return `true` iff the given type is or contains a PodType that can be split into scalars.
+inline static bool containsSplittablePodType(Type t) {
+  return t
+      .walk([](PodType p) {
+    return splittablePod(p) ? WalkResult::interrupt() : WalkResult::skip();
```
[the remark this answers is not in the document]
tim-hoffman: Need to think about the general question of how best to handle types with nested arrays and pods since they are handled by two separate passes. A combined fixpoint option?

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1595 @@
...
+///    - Replace `MemberReadOp` and `MemberWriteOp` targeting the members that were split in step 1
+///      so they instead perform scalar reads and writes from the new members. The transformation is
+///      local to the current op. Therefore, when replacing the `MemberReadOp` a new pod is
+///      created locally and all uses of the `MemberReadOp` are replaced with the new pod Value,
+///      then each scalar member read is followed by scalar write into the new pod. Similarly,
+///      when replacing a `MemberWriteOp`, each element in the pod operand needs a scalar read
+///      from the pod followed by a scalar write to the new member. Making only local changes
+///      keeps this step simple and later steps will optimize.
+///
+///    - Remove optional initialization from `NewPodOp` and instead insert a list of `WritePodOp`
```
iangneal, line 32: Future TODO: seems like it might be nice to break this out into a separate de-sugaring pass (same for optional array initialization)

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1595 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file implements the `-llzk-pod-to-scalar` pass.
+///
+/// The steps of this transformation are as follows:
+///
+/// 0. Scan to find `llzk.nondet` ops that allocate uninitialized pods and replace them with
```
iangneal, line 15: Future TODO: implement a canonicalizer for `llzk.nondet` that will do this for nondet pods/arrays, perhaps.

iangneal (commented):
Just a couple of comments I wanted to record for now while you work through the other comments

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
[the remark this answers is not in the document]
tim-hoffman: The fix is in shared code so the array-to-scalar pass is updated as well. Added regression test for both.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
tim-hoffman: https://github.com/project-llzk/llzk-lib/issues/532
https://github.com/project-llzk/llzk-lib/issues/531

tim-hoffman:
@codex review

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
tim-hoffman: https://github.com/project-llzk/llzk-lib/issues/533

iangneal:
@codex review

include/llzk/Transforms/LLZKConversionUtils.h
```diff
@@ -52,6 +69,120 @@ reserveUniqueFunctionArgName(llvm::StringSet<> &usedNames, llvm::StringRef desir
...
+) {
+  SplitFunctionNameInfo info;
+  info.originalNames.reserve(origTypes.size());
+  info.splitNameSuffixes.reserve(origTypes.size());
+  for (auto [i, type] : llvm::enumerate(origTypes)) {
+    if (std::optional<mlir::StringAttr> nameAttr = getNameAttr(i)) {
+      info.originalNames.push_back(nameAttr->getValue().str());
+      info.existingNames.push_back(nameAttr->getValue().str());
+    } else {
+      info.originalNames.push_back(std::nullopt);
```
0xddom, line 107: The vectors are filled with strings coming from the MLIR context, so I think it should be safe to use `llvm::StringRef` here instead of `std::string` and avoid the copy

include/llzk/Transforms/LLZKConversionUtils.h
```diff
@@ -103,10 +234,17 @@ class FunctionTypeConverter {
     // If the function has a body, ensure the entry block arguments match the function inputs.
     if (mlir::Region *body = op.getCallableRegion()) {
       mlir::Block &entryBlock = body->front();
-      if (!std::cmp_equal(entryBlock.getNumArguments(), newInputs.size())) {
+      bool blockArgsNeedUpdate = !std::cmp_equal(entryBlock.getNumArguments(), newInputs.size());
+      for (unsigned i = 0, e = entryBlock.getNumArguments(); !blockArgsNeedUpdate && i < e; ++i) {
+        blockArgsNeedUpdate = entryBlock.getArgument(i).getType() != newInputs[i];
+      }
```
0xddom: Can't we use `std::any_of` or something similar here?

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+    auto records = pt.getRecords();
+    for (RecordAttr record : records) {
+      collect.push_back(record.getType());
+    }
+    return records.size();
+  } else {
+    collect.push_back(t);
+    return 1;
+  }
+}
```
0xddom, line 190: Is it a pre-condition that the vector is empty?

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+using namespace llzk::pod;
+using namespace llzk::function;
+using namespace llzk::component;
+
+#define DEBUG_TYPE "llzk-pod-to-scalar"
+
+namespace {
+
+/// If the given PodType can be split into scalars (always true for PodType).
+inline static PodType splittablePod(PodType pt) { return pt; }
```
0xddom: Would there be a benefit to returning `false` for pods with only one record? Since those seem to be the target of the splitting

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+    outputOpRef.assign(ValueRange(newOperands));
+  });
+}
+
+/// Register the dialects and operations that remain legal across the conversion-based stages.
+inline static void baseTargetSetup(ConversionTarget &target) {
+  target.addLegalDialect<
+      LLZKDialect, array::ArrayDialect, boolean::BoolDialect, cast::CastDialect,
+      constrain::ConstrainDialect, component::StructDialect, felt::FeltDialect,
+      function::FunctionDialect, global::GlobalDialect, include::IncludeDialect, pod::PODDialect,
```
0xddom, line 281: Why is the pod dialect declared legal here?

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+using RecordChain = ArrayAttr;
+/// new member name and type
+using MemberInfo = std::pair<StringAttr, Type>;
+/// original nested pod record name chain -> split scalar member info
+using LocalMemberReplacementMap = DenseMap<RecordChain, MemberInfo>;
+/// struct -> original pod-type member name -> LocalMemberReplacementMap
+using MemberReplacementMap = DenseMap<StructDefOp, DenseMap<StringAttr, LocalMemberReplacementMap>>;
+
+/// Convert a nested record-name path to an `ArrayAttr` key for the replacement map.
+static ArrayAttr getRecordChainAttr(MLIRContext *ctx, ArrayRef<StringAttr> recordChain) {
```
0xddom: Shouldn't these two lines reference `ArrayAttr` as `RecordChain` since that's the alias name for that context

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+  return ArrayAttr::get(ctx, attrs);
+}
+
+/// Build a flattened struct-member name like `member_outer_inner_leaf`.
+static StringAttr
+getFlattenedMemberName(MLIRContext *ctx, StringAttr memberName, ArrayRef<StringAttr> recordChain) {
+  std::string flatName = memberName.getValue().str();
+  for (StringAttr recordName : recordChain) {
+    flatName += "_" + recordName.getValue().str();
+  }
```
0xddom: Prefer `std::stringstream` or `llvm::raw_string_ostream` over `std::string` and `+=`

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+
+/// Convert a nested record-name path to an `ArrayAttr` key for the replacement map.
+static ArrayAttr getRecordChainAttr(MLIRContext *ctx, ArrayRef<StringAttr> recordChain) {
+  SmallVector<Attribute> attrs;
+  attrs.reserve(recordChain.size());
+  for (StringAttr recordName : recordChain) {
+    attrs.push_back(recordName);
+  }
+  return ArrayAttr::get(ctx, attrs);
+}
```
0xddom: Moreover, are we storing the record chain in any operation at the end? Because using an array attribute will require passing through the uniquing process and allocation of the array in the context's arena. If the elements of the record chain are only meant to be transitive it's better to use something like `SmallVector` instead of `ArrayAttr`

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+    localRepMapRef[getRecordChainAttr(originalMember.getContext(), recordChain)] =
+        std::make_pair(structSymbolTable.insert(newMember), ty);
+    recordChain.pop_back();
+  }
+}
+
+/// Split a pod-typed struct member definition into one scalar member definition per POD record.
+///
+/// The replacement map records the fresh member symbols so later rewrites can retarget
+/// `component.member_read` and `component.member_write` operations to the split members.
```
0xddom:
```suggestion
/// `struct.readm` and `struct.writem` operations to the split members.
```

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+  LogicalResult match(MemberDefOp op) const override { return failure(legal(op)); }
+
+  void
+  rewrite(MemberDefOp op, OpAdaptor adaptor, ConversionPatternRewriter &rewriter) const override {
+    StructDefOp inStruct = op->getParentOfType<StructDefOp>();
+    assert(inStruct);
+    LocalMemberReplacementMap &localRepMapRef = repMapRef[inStruct][op.getSymNameAttr()];
+
+    PodType podTy = dyn_cast<PodType>(adaptor.getType());
+    assert(podTy); // follows from legal() check
```
0xddom:
```suggestion
    PodType podTy = cast<PodType>(adaptor.getType()); // follows from legal() check
```

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+    for (auto [name, init] :
+         llvm::zip_equal(adaptor.getInitializedRecords(), adaptor.getInitialValues())) {
+      // Create the write
+      rewriter.create<WritePodOp>(loc, op.getResult(), llvm::cast<StringAttr>(name), init);
+    }
+    // Remove initializations from `op`
+    rewriter.modifyOpInPlace(op, [&op]() {
+      op.getInitialValuesMutable().clear();
+      op.setInitializedRecordsAttr(ArrayAttr::get(op.getContext(), {})); // DefaultValuedAttr:{}
+    });
```
0xddom, line 436: I'm not sure if it was recommended to use `modifyOpInPlace` while using the conversion engine since it tries to accumulate all conversions during processing and apply them at the end. But that recommendation may have changed from when I heard it so I'm not 100% sure.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+/// Normalize the record name representation used by POD access ops to a plain `StringAttr`.
+inline static StringAttr getRecordNameAsStringAttr(ReadPodOp readOp) {
+  return readOp.getRecordNameAttr().getLeafReference();
+}
+
+/// Normalize the record name representation used by POD access ops to a plain `StringAttr`.
+inline static StringAttr getRecordNameAsStringAttr(WritePodOp writeOp) {
+  return writeOp.getRecordNameAttr().getLeafReference();
+}
+
```
0xddom, line 767: This is why you wanted to change the attribute to a `StringAttr`, right? Makes sense to me if that's the case

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+/// Return whether `op` contains a nested write to `podRef.recordName`.
+static bool hasNestedWriteToRecord(Operation &op, Value podRef, StringAttr recordName) {
+  return op
+      .walk([&](WritePodOp writeOp) {
+    if (writeOp.getOperation() != &op && isSamePodRecord(writeOp, podRef, recordName)) {
+      return WalkResult::interrupt();
+    }
+    return WalkResult::advance();
+  }).wasInterrupted();
+}
```
0xddom, line 783: We could encapsulate this pattern in a high-order function that takes a predicate. Something like the following.

```
template<typename T>
bool anyOp(Operation *op, llvm::function_ref<bool(T)> pred) {
  if (!op) return false;
  return op->walk([&](T t) {
    if (pred(t)) return WalkResult::interrupt();
    return WalkResult::advance();
  }).wasInterrupted();
}
```

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+
+    if (auto writeOp = dyn_cast<WritePodOp>(&op)) {
+      if (isSamePodRecord(writeOp, podRef, recordName)) {
+        replacement = writeOp;
+      }
+      continue;
+    }
+
+    if (hasNestedWriteToRecord(op, podRef, recordName)) {
+      replacement = WritePodOp();
```
0xddom: Wouldn't this create a wrapper to a null operation?

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+    Value carriedPod = writeOp.getPodRef();
+    unsigned podResultIndex = podRes.getResultNumber();
+
+    auto thenYield = dyn_cast<scf::YieldOp>(ifOp.thenBlock()->getTerminator());
+    if (!thenYield || thenYield.getOperand(podResultIndex) != carriedPod) {
+      return;
+    }
+
+    Block *elseBlock = ifOp.getElseRegion().empty() ? nullptr : &ifOp.getElseRegion().front();
+    if (elseBlock) {
```
0xddom:
```suggestion
    if (Block *elseBlock = ifOp.getElseRegion().empty() ? nullptr : &ifOp.getElseRegion().front()) {
```

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+    }
+
+    auto writeOp = dyn_cast_or_null<WritePodOp>(readOp->getPrevNode());
+    if (!writeOp) {
+      return;
+    }
+
+    if (getRecordNameAsStringAttr(writeOp) != getRecordNameAsStringAttr(readOp)) {
+      return;
+    }
```
0xddom: Nit: to reduce the number of lines in the function.
```suggestion
    if (!writeOp || getRecordNameAsStringAttr(writeOp) != getRecordNameAsStringAttr(readOp)) {
      return;
    }
```

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+        return;
+      }
+    }
+
+    replacements.emplace_back(readOp, valueRes);
+  });
+
+  for (auto [readOp, replacement] : replacements) {
+    readOp.getResult().replaceAllUsesWith(replacement);
+    readOp.erase();
```
0xddom: Prefer to do these kind of operations with the `Rewriter`

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+  StringAttr recordName;
+  Type type;
+  WritePodOp thenWrite;
+  WritePodOp elseWrite;
+  Value incomingValue;
+};
+
+/// Find the tracked branch-write slot for `podRef.recordName`.
+static IfWriteSlot *
+lookupSlot(SmallVectorImpl<IfWriteSlot> &slots, Value podRef, StringAttr recordName) {
```
0xddom, line 978:
```suggestion
lookupSlot(ArrayRef<IfWriteSlot> slots, Value podRef, StringAttr recordName) {
```

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+/// Return the `scf.yield` terminator from `block`.
+static scf::YieldOp getYieldOp(Block *block) {
+  assert(block && "expected block");
+  auto yieldOp = dyn_cast<scf::YieldOp>(block->getTerminator());
+  assert(yieldOp && "expected scf.if branch to terminate with scf.yield");
+  return yieldOp;
+}
+
+/// Remove the default terminator from a freshly created SCF block before cloning contents into it.
+static void dropTerminatorIfPresent(Block *block) {
```
0xddom: If `block` cannot be `nullptr` prefer to use `Block&` instead of `Block*`

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+/// Return whether `op` is one of the branch writes that will be recreated after the lifted `if`.
+static bool isLiftedWrite(Operation &op, ArrayRef<IfWriteSlot> slots) {
+  auto writeOp = dyn_cast<WritePodOp>(&op);
+  if (!writeOp) {
+    return false;
+  }
+
+  return llvm::any_of(slots, [&](const IfWriteSlot &slot) {
+    return isSamePodRecord(writeOp, slot.podRef, slot.recordName);
+  });
```
0xddom:
```suggestion
  return writeOp && llvm::any_of(slots, [&](const IfWriteSlot &slot) {
    return isSamePodRecord(writeOp, slot.podRef, slot.recordName);
  });
```

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+  }
+
+  for (auto it = srcBlock->begin(), end = srcBlock->end(); it != end;) {
+    Operation &op = *it++;
+    if (op.hasTrait<OpTrait::IsTerminator>() || isLiftedWrite(op, slots)) {
+      continue;
+    }
+    op.moveBefore(destBlock, destBlock->end());
+  }
+}
```
0xddom, line 1097: Consistency nit: In this function `srcBlock` is checked for null (that can be removed by using `Block&`) but `destBlock` is not

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+
+static const LoopPodSlot *
+lookupLoopSlot(ArrayRef<LoopPodSlot> slots, Value podRef, StringAttr recordName) {
+  for (const LoopPodSlot &slot : slots) {
+    if (slot.podRef == podRef && slot.recordName == recordName) {
+      return &slot;
+    }
+  }
+  return nullptr;
+}
```
0xddom: These functions are duplicated

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+      return idx;
+    }
+  }
+  return std::nullopt;
+}
+
+/// Collect direct POD reads and writes that cross the loop boundary and therefore need an
+/// explicit loop-carried scalar value when lifted.
+static void
+collectDirectLoopPodSlots(Block *block, Operation *ancestor, SmallVectorImpl<LoopPodSlot> &slots) {
```
0xddom: Nit: Use `Block&`

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -0,0 +1,1708 @@
...
+  for (auto [idx, oldArg] : llvm::enumerate(forOp.getRegionIterArgs())) {
+    mapping.map(oldArg, newFor.getRegionIterArg(idx));
+  }
+
+  SmallVector<Value> slotValues;
+  slotValues.reserve(slots.size());
+  for (auto [idx, slot] : llvm::enumerate(slots)) {
+    (void)slot;
+    slotValues.push_back(newFor.getRegionIterArg(forOp.getNumRegionIterArgs() + idx));
+  }
```
0xddom: You could use `llvm::map_to_vector` for this one.

0xddom (commented):
Mostly nits and comments. I think there are some functions that could be simplified or abstracted. Other than that, the general algorithm LGTM, tho the third step was a bit hard to follow

test/Transforms/InlineStructs/circom_subcmps5b.llzk
0xddom: Yeah, the way I wrote the IR generation was conducive to getting DCEd. So nice seeing the tests working as I had in mind.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
tim-hoffman: No, it's building the new parameter type list by visiting each existing type and appending one or more types.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
tim-hoffman: Good intuition but changing it caused single-record pod type to be preserved.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
tim-hoffman: This is used in the early steps and pod would not be fully removed until after the loop in `runOnOperation()` completes.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
tim-hoffman: Original reason was @iangneal had to specifically filter out those ops in a symbol-related analysis because they're not referencing true symbols.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
tim-hoffman: The `ArrayRef` value type is `const` but the value returned is ultimately mutated.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
tim-hoffman: It's the constness of `ArrayRef` again. I think it's all just SmalVector that's created on the stack so I can probably just reduce to the first one.

tim-hoffman: Actually, the second is just use to check if it exists so I'll change it to return `bool`.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
tim-hoffman: It's safe, just modification that doesn't use the `ConversionPatternRewriter` is unsafe.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -775,37 +776,23 @@ inline static bool isSamePodRecord(WritePodOp writeOp, Value podRef, StringAttr
...
 /// Return whether `op` contains any read from `podRef.recordName`.
 static bool hasReadFromRecord(Operation &op, Value podRef, StringAttr recordName) {
-  return op
-      .walk([&](ReadPodOp readOp) {
-    if (isSamePodRecord(readOp, podRef, recordName)) {
-      return WalkResult::interrupt();
-    }
-    return WalkResult::advance();
-  }).wasInterrupted();
+  return walkContainsMatch<ReadPodOp>(op, [&](ReadPodOp readOp) {
```
0xddom, line 787: There's a way of getting rid of `<ReadPodOp>`, but it requires some template black magic that I don't remember very well how it works. We can do it some other time

tim-hoffman: I feel like `TypeSwitch.Case` could be using that black magic

tim-hoffman:
@codex review

iangneal (approved):
LGTM!

PR #516, opened 2026-06-03
add op-restricted SROA and mem2reg passes

tim-hoffman:
In preparation for adding the interfaces for these passes to `pod` dialect ops (which would cause a single run of the builtin passes to process all ops with the pass-related interface), create op-restricted SROA and mem2reg passes. This allows `array` destructuring and `pod` destructuring to be run as independent passes (likewise for `pod` when that is updated).


iangneal (approved):
LGTM!

PR #518, opened 2026-06-04
Prime field fix

iangneal:
- Changed the `bn128/bn254` to use the correct scalar field prime (was using the base field prime instead, causing discrepancies when comparing to circom witness generator output)
- Added `grumpkin` field definition (the `grumpkin` scalar field prime is the same as the `bn128`/`bn254` base field prime, which was likely the original source of confusion)

Constants now match the definitions in the circom repo (links added in documentation).


unittests/Analysis/DynamicAPIntTests.cpp
```diff
@@ -71,7 +71,7 @@ struct DynamicAPIntStringTest : public testing::TestWithParam<std::string> {
         "2147483647",           // mersenne31
         "2130706433",           // koalabear
         "18446744069414584321", // goldilocks
-        "21888242871839275222246405745257275088696311157297823662689037894645226208583", // bn254
+        "21888242871839275222246405745257275088548364400416034343698204186575808495617", // bn254
```
tim-hoffman, line 74: I'd rather test grumpkin here since it's larger

tim-hoffman (approved):
LGTM, the one comment I gave doesn't really matter since int64::max is also in the test set

PR #522, opened 2026-06-04
Restore insert point properly

0xddom:
Second round of this because I missed one thing in #521. The CAPI side of the `Block::iterator` needs to be a null pointer, however the value that indicates "the end of the block" is not a null pointer but `Block::end`, that is probably some sort of sentinel value (`llvm::ilist` shenanigans). So this fix reconstructs the insert point properly, using `Block::end` when the CAPI point is a null operation.

I also added a new test that ensures the restoration is good by calling `mlirOpBuilderInsert` since that's the point where it will try use the insert point that was previously restored.


tim-hoffman (approved):
LGTM

PR #535, opened 2026-06-09
Prevent CSE from merging array.new allocations

### LINKED ISSUE #509: `array.new` CSE changes witness output by merging fresh mutable arrays
## Summary

Running `--cse` on a program with two identical `array.new` operations can merge the two array allocations. If one array is written and the other is later read, the optimized program reads from the written array instead of the independent fresh array, changing the public witness output.

On current `main` (`04708f582ed1c0c61e8407e43c88cd38d46f2fc1`), the same program and input produce:

```text
before --cse: {"out":"3"}
after  --cse: {"out":"9"}
```

The repro was also run in GitHub Actions here:

https://github.com/Kuhai9801/llzk-lib/actions/runs/26800490487

## Reproduction

Save this as `array_new_cse_witness.llzk`:

```llzk
module attributes {llzk.lang = "llzk", llzk.main = !struct.type<@Main>} {
  struct.def @Main {
    struct.member @out : !felt.type<"babybear"> {llzk.pub}
    function.def @compute(
      %a: !felt.type<"babybear"> {function.arg_name = "a"},
      %b: !felt.type<"babybear"> {function.arg_name = "b"}
    ) -> !struct.type<@Main> {
      %self = struct.new : !struct.type<@Main>
      %i = arith.constant 0 : index
      %x = array.new %a : !array.type<1 x !felt.type<"babybear">>
      %y = array.new %a : !array.type<1 x !felt.type<"babybear">>
      array.write %x[%i] = %b : !array.type<1 x !felt.type<"babybear">>, !felt.type<"babybear">
      %r = array.read %y[%i] : !array.type<1 x !felt.type<"babybear">>, !felt.type<"babybear">
      struct.writem %self[@out] = %r : !struct.type<@Main>, !felt.type<"babybear">
      function.return %self : !struct.type<@Main>
    }
    function.def @constrain(
      %self: !struct.type<@Main>,
      %a: !felt.type<"babybear">,
      %b: !felt.type<"babybear">
    ) {
      function.return
    }
  }
}
```

Save this as `input.json`:

```json
{"a": 3, "b": 9}
```

Run:

```sh
llzk-witgen array_new_cse_witness.llzk --inputs input.json
llzk-opt array_new_cse_witness.llzk --cse > array_new_cse_witness.after-cse.llzk
llzk-witgen array_new_cse_witness.after-cse.llzk --inputs input.json
```

## Observed Behavior

Before CSE:

```json
{
  "out": "3"
}
```

After CSE:

```json
{
  "out": "9"
}
```

The optimized IR contains one array allocation shared by the write and read:

```mlir
%array = array.new %arg0 : <1 x !felt.type<"babybear">>
array.write %array[%c0] = %arg1 : <1 x !felt.type<"babybear">>, !felt.type<"babybear">
%0 = array.read %array[%c0] : <1 x !felt.type<"babybear">>, !felt.type<"babybear">
struct.writem %self[@out] = %0 : <@Main>, !felt.type<"babybear">
```

## Expected Behavior

The two `array.new %a` operations should remain distinct allocations. Writing `%b` into `%x` should not affect a read from `%y`, so the public output should remain `"3"` after CSE.

## Source Context

`array.new` is declared with the `Pure` trait:

https://github.com/project-llzk/llzk-lib/blob/04708f582ed1c0c61e8407e43c88cd38d46f2fc1/include/llzk/Dialect/Array/IR/Ops.td#L103-L113

The same op is described as creating a new array:

https://github.com/project-llzk/llzk-lib/blob/04708f582ed1c0c61e8407e43c88cd38d46f2fc1/include/llzk/Dialect/Array/IR/Ops.td#L114-L129

`array.read` observes an array position:

https://github.com/project-llzk/llzk-lib/blob/04708f582ed1c0c61e8407e43c88cd38d46f2fc1/include/llzk/Dialect/Array/IR/Ops.td#L240-L250

`array.write` mutates an array position:

https://github.com/project-llzk/llzk-lib/blob/04708f582ed1c0c61e8407e43c88cd38d46f2fc1/include/llzk/Dialect/Array/IR/Ops.td#L274-L288

`array.new` also implements allocation-like behavior through destructuring and promotion hooks:

https://github.com/project-llzk/llzk-lib/blob/04708f582ed1c0c61e8407e43c88cd38d46f2fc1/lib/Dialect/Array/IR/Ops.cpp#L123-L132

https://github.com/project-llzk/llzk-lib/blob/04708f582ed1c0c61e8407e43c88cd38d46f2fc1/lib/Dialect/Array/IR/Ops.cpp#L175-L191

`array.write` is also modeled as a store to a memory slot:

https://github.com/project-llzk/llzk-lib/blob/04708f582ed1c0c61e8407e43c88cd38d46f2fc1/lib/Dialect/Array/IR/Ops.cpp#L362-L379

Given those semantics, treating two identical `array.new` operations as freely CSE-able can alias distinct mutable arrays and change witness output.

## Possible Fix Direction

`array.new` should not be CSE-able when it creates a fresh mutable allocation. One possible direction is to remove the `Pure` trait from `array.new` or replace it with memory-effect modeling that distinguishes fresh allocations from ordinary pure value construction.

1sgtpepper:
## Summary

Fixes #509 by removing `Pure` from `array.new`, which creates a fresh mutable allocation and must not be merged by CSE.

## Changes

- Remove `Pure` from `LLZK_CreateArrayOp`.
- Add a CSE regression for distinct `array.new` allocations.
- Preserve cleanup of unused array allocations after promotion/flattening.
- Document fresh allocation semantics.
- Add an unreleased changelog entry.

## Validation

- `git diff --check`
- CSE regression: `llzk-opt --pass-pipeline='builtin.module(cse, canonicalize)' test/Dialect/Array/array_new_cse_preservation.llzk | FileCheck --enable-var-scope test/Dialect/Array/array_new_cse_preservation.llzk`
- GitHub Actions: changelog validation, docs build, formatting check, Nix checks, and GCC/Clang test jobs


lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
```diff
@@ -2525,7 +2526,9 @@ class FlatteningPass : public llzk::polymorphic::impl::FlatteningPassBase<Flatte
     if (failed(runPipeline(universalCleanup, modOp))) {
       return failure();
     }
-    return success();
+    OpPassManager finalCleanup(ModuleOp::getOperationName());
+    finalCleanup.addPass(createCanonicalizerPass());
+    return runPipeline(finalCleanup, modOp);
```
0xddom: I don't think is really that necessary that we run the canonicalizer here. What do you reckon @tim-hoffman ?

tim-hoffman: Agreed. In general, I'd rather not bundle it with some other pass since its effects are so far reaching; let the user specify it if necessary.

include/llzk/Dialect/Array/IR/Ops.td
```diff
@@ -103,7 +103,7 @@ class RangedArrayAccessOp<string mnemonic, bit isRead, list<Trait> traits = []>
 def LLZK_CreateArrayOp
     : ArrayDialectOp<
           "new",
-          [Pure, AttrSizedOperandSegments, VerifySizesForMultiAffineOps<1>,
+          [AttrSizedOperandSegments, VerifySizesForMultiAffineOps<1>,
```
iangneal: We might want to explicitly add `MemoryEffects<[MemAlloc<DefaultResource>]>`

lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
tim-hoffman: Same in `ArrayToScalarPass`. The addition of canonicalization there makes sense to me (since removal of Pure affects RDV) but I would prefer a different approach. However, it's not obvious to me how the removal of Pure affected `FlatteningPass`.

include/llzk/Dialect/Array/IR/Ops.td
tim-hoffman: @iangneal I'm still annoyed that upstream MLIR has no way of annotating a value that can be removed by RDV is unused but cannot be combined by CSE. Perhaps we should add a custom pass to remove `MemAlloc` values that are unused, maybe with a custom resource that indicates there's no externally observable effect from that allocation.

iangneal: Oh that sounds like a nice solution.

tim-hoffman: `eraseDeadAllocAndStores()` in `mlir/lib/Dialect/MemRef/Utils/MemRefUtils.cpp` could be copied and adapted for our pass

include/llzk/Transforms/SpecializedMemoryPasses.h
```diff
@@ -1,4 +1,4 @@
-//===-- SpecializedMemoryPasses.h - Targeted SROA / mem2reg -----*- C++ -*-===//
+//===-- SpecializedMemoryPasses.h - Targeted memory passes -------*- C++ -*-===//
```
tim-hoffman: nit
```suggestion
//===-- SpecializedMemoryPasses.h - Targeted memory passes ------*- C++ -*-===//
```

1sgtpepper:
Updated to use `MemAlloc` and a targeted unused-allocation cleanup instead of the hidden canonicalizer.

include/llzk/Transforms/SpecializedMemoryPasses.h
```diff
@@ -127,4 +130,162 @@ std::unique_ptr<SpecializedMem2Reg<AllocOpTy>> createSpecializedMem2RegPass() {
   return std::make_unique<SpecializedMem2Reg<AllocOpTy>>();
 }
 
+/// A pass that erases allocator ops of type \p AllocOpTy when their remaining uses are removable
+/// stores. Use this only for allocator ops whose allocation has no externally observable effect
+/// unless its stored values are read or escaped.
```
tim-hoffman: What does "escaped" mean here?

include/llzk/Transforms/SpecializedMemoryPasses.h
```diff
@@ -127,4 +130,162 @@ std::unique_ptr<SpecializedMem2Reg<AllocOpTy>> createSpecializedMem2RegPass() {
...
+      changed |= changedThisIteration;
+    } while (changedThisIteration);
+
+    if (!changed) {
+      this->markAllAnalysesPreserved();
+    }
+  }
+
+private:
+  static bool collectRemovableUsers(
```
tim-hoffman: please document

include/llzk/Transforms/SpecializedMemoryPasses.h
```diff
@@ -127,4 +130,162 @@ std::unique_ptr<SpecializedMem2Reg<AllocOpTy>> createSpecializedMem2RegPass() {
...
+      if (!memOp || !canRemoveUse(memOp, use, slots, dataLayout)) {
+        return false;
+      }
+      usersToErase.push_back(user);
+    }
+
+    return true;
+  }
+
+  static bool canRemoveUse(
```
tim-hoffman: please document

include/llzk/Transforms/SpecializedMemoryPasses.h
```diff
@@ -127,4 +130,162 @@ std::unique_ptr<SpecializedMem2Reg<AllocOpTy>> createSpecializedMem2RegPass() {
...
+
+    return llvm::any_of(slots, [&](mlir::MemorySlot slot) {
+      mlir::SmallVector<mlir::OpOperand *> newBlockingUses;
+      return memOp.storesTo(slot) && !memOp.loadsFrom(slot) &&
+             memOp.canUsesBeRemoved(slot, blockingUses, newBlockingUses, dataLayout) &&
+             newBlockingUses.empty();
+    });
+  }
+
+  static void collectOperandDefiningOps(
```
tim-hoffman: please document

include/llzk/Transforms/SpecializedMemoryPasses.h
```diff
@@ -127,4 +130,162 @@ std::unique_ptr<SpecializedMem2Reg<AllocOpTy>> createSpecializedMem2RegPass() {
...
+        mlir::Operation *definingOp = operand.getDefiningOp();
+        if (definingOp && !opsBeingErased.contains(definingOp) &&
+            seenDefs.insert(definingOp).second) {
+          maybeDeadDefs.push_back(definingOp);
+        }
+      }
+    }
+  }
+
+  static bool eraseTriviallyDeadDefs(mlir::SmallVectorImpl<mlir::Operation *> &maybeDeadDefs) {
```
tim-hoffman: please document

include/llzk/Transforms/SpecializedMemoryPasses.h
```diff
@@ -127,4 +130,162 @@ std::unique_ptr<SpecializedMem2Reg<AllocOpTy>> createSpecializedMem2RegPass() {
   return std::make_unique<SpecializedMem2Reg<AllocOpTy>>();
 }
 
+/// A pass that erases allocator ops of type \p AllocOpTy when their remaining uses are removable
```
tim-hoffman: I've only scanned through so far but it looks ops besides allocators are removed (some users of the allocated value) so the documentation should be updated.

lib/Dialect/Array/Transforms/ArrayToScalarPass.cpp
```diff
@@ -950,6 +953,10 @@ class ArrayToScalarPass : public llzk::array::impl::ArrayToScalarPassBase<ArrayT
     nestedPM.addPass(createSpecializedMem2RegPass<CreateArrayOp>());
     // Cleanup SSA values made dead by the transformations
     nestedPM.addPass(createRemoveDeadValuesPass());
```
tim-hoffman: Could we remove this first RDV or does it reveal ops that can be removed by the unused allocations pass?

include/llzk/Dialect/Array/IR/Ops.td
```diff
@@ -103,7 +104,8 @@ class RangedArrayAccessOp<string mnemonic, bit isRead, list<Trait> traits = []>
 def LLZK_CreateArrayOp
     : ArrayDialectOp<
           "new",
-          [Pure, AttrSizedOperandSegments, VerifySizesForMultiAffineOps<1>,
+          [MemoryEffects<[MemAlloc<DefaultResource>]>, AttrSizedOperandSegments,
```
tim-hoffman: I think it would be best to create a new resource `DiscardableAllocationResource` to use instead of `DefaultResource` and then have the `SpecializedRemoveUnusedAllocations` require the given op type has  the `MemAlloc<DiscardableAllocationResource>` memory effect.

include/llzk/Transforms/SpecializedMemoryPasses.h
```diff
@@ -127,4 +130,169 @@ std::unique_ptr<SpecializedMem2Reg<AllocOpTy>> createSpecializedMem2RegPass() {
   return std::make_unique<SpecializedMem2Reg<AllocOpTy>>();
 }
 
+/// Erases allocators of type \p AllocOpTy whose remaining direct users are write-only memory ops.
```
tim-hoffman:
```suggestion
/// Erases allocators of type \p AllocOpTy that are only stored to and never loaded from.
```

tim-hoffman (commented):
A couple more things I noticed while reviewing the pass def

tim-hoffman:
@codex review

include/llzk/Transforms/SpecializedMemoryPasses.h
```diff
@@ -127,4 +149,176 @@ std::unique_ptr<SpecializedMem2Reg<AllocOpTy>> createSpecializedMem2RegPass() {
...
+      AllocOpTy allocator, mlir::SmallVectorImpl<mlir::Operation *> &usersToErase,
+      const mlir::DataLayout &dataLayout
+  ) {
+    if (allocator->use_empty()) {
+      return true;
+    }
+
+    mlir::SmallVector<mlir::MemorySlot> slots = allocator.getPromotableSlots();
+    if (slots.empty()) {
+      return false;
```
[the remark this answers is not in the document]
tim-hoffman: I was wondering if something like this would come up because the "promotable" interfaces are used here instead of a specialized op interface.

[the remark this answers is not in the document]
tim-hoffman: Maybe with some methods like this:
```
    InterfaceMethod<
      [{Returns true if this op reads from the allocation represented by `ptr`.}],
      "bool", "loadsFromDiscardableAllocation",
      (ins "::mlir::Value":$ptr)>,

    InterfaceMethod<
      [{Returns true if this op writes to the allocation represented by `ptr`.}],
      "bool", "storesToDiscardableAllocation",
      (ins "::mlir::Value":$ptr)>,

    InterfaceMethod<
      [{Returns true if this op can be erased as a dead store to `ptr`.}],
      "bool", "canEraseAsDeadStoreTo",
      (ins "::mlir::Value":$ptr,
           "const ::mlir::DataLayout &":$dataLayout)>
```

include/llzk/Dialect/Array/IR/Ops.td
```diff
@@ -103,7 +107,8 @@ class RangedArrayAccessOp<string mnemonic, bit isRead, list<Trait> traits = []>
 def LLZK_CreateArrayOp
     : ArrayDialectOp<
           "new",
-          [Pure, AttrSizedOperandSegments, VerifySizesForMultiAffineOps<1>,
+          [MemoryEffects<[MemAlloc<DiscardableAllocationResource>]>,
```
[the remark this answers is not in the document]
tim-hoffman: this is unimportant

lib/Dialect/Array/Transforms/ArrayToScalarPass.cpp
```diff
@@ -948,7 +951,11 @@ class ArrayToScalarPass : public llzk::array::impl::ArrayToScalarPassBase<ArrayT
     nestedPM.addPass(createSpecializedSROAPass<CreateArrayOp>());
     // The mem2reg pass converts all of the size-1 array allocation and access into SSA values.
     nestedPM.addPass(createSpecializedMem2RegPass<CreateArrayOp>());
-    // Cleanup SSA values made dead by the transformations
+    // Cleanup allocations made dead by memory promotion.
+    nestedPM.addPass(
+        createSpecializedRemoveUnusedAllocationsPass<CreateArrayOp, DiscardableAllocationResource>()
+    );
+    // Cleanup SSA values made dead by removing allocations and writes.
     nestedPM.addPass(createRemoveDeadValuesPass());
```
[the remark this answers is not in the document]
tim-hoffman: This is in pre-existing code so do not fix in this PR although it could be a valid concern.

include/llzk/Dialect/Array/IR/Ops.td
```diff
@@ -86,6 +88,23 @@ class ScalarArrayAccessOp<string mnemonic, bit isRead, list<Trait> traits = []>
...
+    /// Required by DiscardableAllocationAccessorOpInterface / unused allocation cleanup.
+    bool $cppClass::storesToDiscardableAllocation(::mlir::Value ptr) {
+      return }]#!if(isRead, "false", "getArrRef() == ptr")#[{;
+    }
+
+    /// Required by DiscardableAllocationAccessorOpInterface / unused allocation cleanup.
+    bool $cppClass::canEraseAsDeadStoreTo(::mlir::Value ptr,
+                                          const ::mlir::DataLayout & /*dataLayout*/) {
+      }]#!if(isRead, "return false",
+             "return getArrRef() == ptr && getRvalue() != ptr")#[{;
```
tim-hoffman, line 105: nit
```suggestion
      return }]#!if(isRead, "false", "getArrRef() == ptr && getRvalue() != ptr")#[{;
```

lib/Dialect/Array/Transforms/ArrayToScalarPass.cpp
```diff
@@ -953,7 +953,9 @@ class ArrayToScalarPass : public llzk::array::impl::ArrayToScalarPassBase<ArrayT
     nestedPM.addPass(createSpecializedMem2RegPass<CreateArrayOp>());
     // Cleanup allocations made dead by memory promotion.
     nestedPM.addPass(
-        createSpecializedRemoveUnusedAllocationsPass<CreateArrayOp, DiscardableAllocationResource>()
+        createSpecializedRemoveUnusedAllocationsPass<
+            CreateArrayOp, DiscardableAllocationResource,
+            DiscardableAllocationAccessorOpInterface>()
```
tim-hoffman: `DiscardableAllocationResource` and `DiscardableAllocationAccessorOpInterface` should always be used in the pass rather than given as template parameters because those are the only ones that will be used. They were created specifically for this pass.

include/llzk/Dialect/Array/IR/OpInterfaces.td
```diff
@@ -83,4 +83,28 @@ def ArrayAccessOpInterface
   }];
 }
 
+def DiscardableAllocationAccessorOpInterface
+    : OpInterface<"DiscardableAllocationAccessorOpInterface"> {
+  let description = [{
+    Interface for operations that can classify and erase their use of a
+    discardable allocation without requiring that allocation to be promotable.
```
tim-hoffman:
```suggestion
    Provides the necessary information to the `SpecializedRemoveUnusedAllocations` pass
    about operations that load or store values from the result of an allocator op marked with
    the `MemAlloc<DiscardableAllocationResource>` memory effect so unused allocators
    and stores can be removed.
```

include/llzk/Transforms/SpecializedMemoryPasses.h
```diff
@@ -151,13 +151,15 @@ std::unique_ptr<SpecializedMem2Reg<AllocOpTy>> createSpecializedMem2RegPass() {
...
 ///
-/// The allocator must have a `MemAlloc<ResourceTy>` effect. Any read, terminator use,
-/// non-promotable user, or memory op that cannot remove its use keeps the allocator and all users
-/// intact.
-template <typename AllocOpTy, typename ResourceTy>
+/// The allocator must have a `MemAlloc<ResourceTy>` effect and exactly one result. Any read,
+/// terminator use, non-discardable-accessor user, or accessor that cannot be erased as a dead
+/// store keeps the allocator and all users intact.
+template <typename AllocOpTy, typename ResourceTy, typename DiscardableAccessorOpInterfaceTy>
 struct SpecializedRemoveUnusedAllocations
```
tim-hoffman: I think the name `RemoveUnusedDiscardableAllocations` would be more appropriate

unittests/IR/SpecializedMemoryPassesTests.cpp
```diff
@@ -85,4 +123,25 @@ TEST_F(SpecializedRemoveUnusedAllocationsTest, ErasesUnreadDiscardableAllocation
...
+  OwningOpRef<ModuleOp> module = parseModuleWithReadAfterArrayWrite();
+  ASSERT_TRUE(module);
+
+  runCleanup<DiscardableAllocationResource>(*module);
+
+  EXPECT_EQ(countOps<CreateArrayOp>(*module), 1u);
+  EXPECT_EQ(countOps<WriteArrayOp>(*module), 1u);
+  EXPECT_EQ(countOps<ReadArrayOp>(*module), 1u);
+}
+
```
tim-hoffman: I think it would be better to refactor the new pass to be externally accessible by defining it in `include/llzk/Transforms/LLZKTransformationPasses.td`, etc. and then make these into lit tests instead of unit tests. Additionally, `test/Transforms/Flattening/remove_write_only_arrays.llzk` should just use that pass instead of `llzk-flatten` and be moved into a new directory containing tests for this new pass.

tim-hoffman (changes requested):
Ok, this is looking good. There are a few more changes I would like to see to get it just right.

unittests/IR/SpecializedMemoryPassesTests.cpp
tim-hoffman: Update the changelog as well to note addition of this new pass

include/llzk/Transforms/LLZKTransformationPasses.td
```diff
@@ -52,6 +52,17 @@ def UnusedDeclarationEliminationPass
   ];
 }
 
+def RemoveUnusedDiscardableAllocationsPass
+    : LLZKPass<"llzk-remove-unused-discardable-allocations"> {
+  let summary = "Remove unread discardable array allocations and their dead stores";
+  let description = [{
+    Remove `array.new` ops marked with `MemAlloc<DiscardableAllocationResource>` when the
```
tim-hoffman: I apologize if I have not made this clear but the pass must not be specific to array because I intend to use this on the `pod` dialect as well in the future.
```suggestion
    Remove ops marked with `MemAlloc<DiscardableAllocationResource>` when the
```

lib/Transforms/LLZKRemoveUnusedDiscardableAllocationsPass.cpp
```diff
@@ -0,0 +1,196 @@
...
+} // namespace llzk
+
+using namespace mlir;
+using namespace llzk;
+using namespace llzk::array;
+
+namespace {
+
+/// Returns true when the allocator is explicitly marked as safe for discardable-allocation cleanup.
+static bool hasDiscardableAllocationEffect(CreateArrayOp allocator) {
```
tim-hoffman: This has regressed to be specific to `CreateArrayOp` again. You will have to add a pass option that can accept a string like "array.new" or "pod.new" and verify that the op has the memory effect we added, else give an error. The programatic callers added by this PR should use `CreateArrayOp::getOperationName()` for the option.

tim-hoffman (changes requested):

1sgtpepper:
Does this make sense? I think you want the pass itself to stay allocator-agnostic, with `array.new` just passed in by the current callers.

I updated it so the pass takes an `allocator-op` option, checks for `MemAlloc<DiscardableAllocationResource>`, and the array callers pass `CreateArrayOp::getOperationName()`.

tim-hoffman:
> Does this make sense? I think you want the pass itself to stay allocator-agnostic, with `array.new` just passed in by the current callers.
> 
> I updated it so the pass takes an `allocator-op` option, checks for `MemAlloc<DiscardableAllocationResource>`, and the array callers pass `CreateArrayOp::getOperationName()`.

Yes, that sounds right. Thanks! I'll review again today.

doc/doxygen/03_syntax.md
```diff
@@ -43,6 +43,8 @@ module attributes {llzk.lang = "circom"} {
 
 LLZK supports arrays where the element type is not truly homogeneous, specifically when a templated `struct.type` is used with an `affine_map` parameter. For example, the type `!array.type<10 x !struct.type<@X<[affine_map<(i)[] -> (i*5)>]>>>` contains instances of the struct `@X` instantiated with different parameter values per `affine_map<(i)[] -> (i*5)>`. Use of this type can be seen in [circom_example_2.llzk](\ref test/FrontendLang/Circom/circom_example_2.llzk). If the circuit is ultimately instantiated and flattened, the array will have to be split into scalar values since the instantiated struct type of each element is different.
 
+Each `array.new` operation creates a fresh mutable array allocation. Two identical `array.new` operations are not interchangeable when either result may be read or written.
+
```
tim-hoffman: Move to "semantic rules" section below. I think it makes more sense there.

include/llzk/Transforms/SpecializedMemoryPasses.h
```diff
@@ -21,6 +19,7 @@
 #include <mlir/IR/Builders.h>
 #include <mlir/IR/Dominance.h>
 #include <mlir/Interfaces/MemorySlotInterfaces.h>
+#include <mlir/Interfaces/SideEffectInterfaces.h>
 #include <mlir/Pass/Pass.h>
 #include <mlir/Transforms/Mem2Reg.h>
 #include <mlir/Transforms/SROA.h>
```
tim-hoffman, line 26: revert all changes here since the new pass moved elsewhere

lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
```diff
@@ -25,6 +25,8 @@
 #include "llzk/Dialect/Polymorphic/Transforms/TransformationPasses.h"
 #include "llzk/Dialect/String/IR/Dialect.h"
 #include "llzk/Dialect/Struct/IR/Ops.h"
+#include "llzk/Transforms/LLZKTransformationPasses.h"
+#include "llzk/Transforms/SpecializedMemoryPasses.h"
```
tim-hoffman: this include is not used anymore, right?
```suggestion
```

include/llzk/Dialect/Array/IR/Ops.td
tim-hoffman: Can you make this change (unless I messed up the syntax and it's not valid?)

tim-hoffman (changes requested):
looks good, just a few minor changes and then I can approve

tim-hoffman (approved):
Thanks for your contribution!

PR #536, opened 2026-06-10
Allow non-native field ops in `verif.contract`

tim-hoffman:


include/llzk/Dialect/Shared/OpHelpers.h
```diff
@@ -62,7 +104,7 @@ template <typename... Ancestors> struct HasAncestor {
   // NOLINTNEXTLINE(bugprone-crtp-constructor-accessibility)
   struct Impl : public mlir::OpTrait::TraitBase<ConcreteType, Impl> {
     static mlir::LogicalResult verifyRegionTrait(mlir::Operation *op) {
-      if (op->getParentOfType<Ancestors...>()) {
+      if (hasParentThatIsa<Ancestors...>(op)) {
```
0xddom, line 107: `Operation::getParentOfType` already does what the replacement is trying to do. If you look at the implementation is practically the same code as `hasParentThatIsa`

lib/Dialect/Function/IR/OpTraits.cpp
```diff
@@ -46,15 +48,17 @@ LogicalResult verifyWitnessGenTraitImpl(Operation *op) {
 }
 
 LogicalResult verifyNotFieldNativeTraitImpl(Operation *op) {
-  if (op->getParentOfType<llzk::polymorphic::TemplateExprOp>()) {
+  // Special cases where these ops are allowed
+  if (hasParentThatIsa<llzk::polymorphic::TemplateExprOp, llzk::verif::ContractOp>(op)) {
```
0xddom: Didn't we said it would be a blacklist? If so then shouldn't this line be something like the folllwing?

```
if (!op->getParentOfType<llzk::function::FuncDefOp>()) {
```

0xddom (changes requested):
There's some redundant code and I'm not sure the check is doing what we said it should

include/llzk/Dialect/Shared/OpHelpers.h
tim-hoffman: The original won't compile if the parameter pack has more than one class listed

lib/Dialect/Function/IR/OpTraits.cpp
tim-hoffman: My earlier comment was only in regards to what the error message should say but I do see what you're saying now.

include/llzk/Dialect/Shared/OpHelpers.h
0xddom: Turns out the version that allows more than one parameter is post LLVM 20. Forgot that the online docs don't necessarily match what's available to us

0xddom (approved):
LGTM!

PR #540, opened 2026-06-12
Loop invariant operations

0xddom:
This PR adds operations for expressing loop invariants.

Invariants are defined with the `verif.invariant` operation and the other operations added in this PR 
are meant to be used within the body of that operation.

Also introduces two interfaces for handling contract and invariant targets. These interfaces abstract the exact operation that is 
targeted by either a contract or an invariant. The contract target interface has been implemented for `function.def` and `struct.def`
while the invariant target interface has been implemented for `scf.while` and `scf.for`.


test/Dialect/Verif/contracts_fail.llzk
```diff
@@ -1156,3 +1156,4 @@ module attributes {llzk.lang} {
     verif.ensure_compute %1
   }
 }
+
```
iangneal:
```suggestion
```

test/Dialect/Verif/invariants_fail.llzk
iangneal, line 1: We should add some tests here to check for mismatched loop args and init args.

test/Dialect/Verif/invariants_pass.llzk
iangneal, line 1: We should add some tests here for loops with init args

include/llzk/Dialect/Verif/IR/OpInterfaces.td
```diff
@@ -48,4 +49,54 @@ def PostconditionOpInterface
...
+
+    Targets are referenced by name so implementations of this interface must also implement the `Symbol` interface.
+  }];
+  let cppNamespace = "::llzk::verif";
+
+  let methods = [InterfaceMethod<
+      [{Gets the ops that can be targeted by invariant ops.}],
+      "::llvm::SmallVector<::llzk::verif::InvariantTargetOpInterface>",
+      "getLoops", (ins), [{
+                        ::mlir::SmallVector<::llzk::verif::InvariantTargetOpInterface> targets;
```
iangneal: I think this can be replaced with `walkCollect`

include/llzk/Dialect/Verif/IR/Ops.td
```diff
@@ -481,4 +491,124 @@ def IncludeOp
...
+    // https://github.com/llvm/llvm-project/blob/f325e4b2d836d6e65a4d0cf3efc6b0996ccf3765/mlir/lib/Dialect/ControlFlow/IR/ControlFlowOps.cpp#L92-L97
+    void $cppClass::getEffects(
+        ::mlir::SmallVectorImpl<::mlir::SideEffects::EffectInstance<::mlir::MemoryEffects::Effect>> &effects
+    ) {
+      effects.emplace_back(::mlir::MemoryEffects::Write::get());
+    }
+  }];
+}
+
+def IncreasesOp : InvariantInnerOp<"increases"> {}
```
iangneal, line 598: I didn't see any tests with these inner ops added, we should add some.

iangneal (changes requested):

1sgtpepper:
I had one broader concern while reading through this.

The feature direction makes sense, but `verif.invariant` still feels under-specified as an IR construct. The op defines a loop label, argument types, and a single block, but the verifier currently seems to only resolve the loop label and check argument type compatibility. I do not see verifier rules that define what a valid invariant body must contain or what shape later passes can rely on. See the [`verif.invariant` definition](https://github.com/project-llzk/llzk-lib/blob/638b6f48fe6643ec3e17ac6927d2f48fac96f663/include/llzk/Dialect/Verif/IR/Ops.td#L498-L524) and the [invariant verifier](https://github.com/project-llzk/llzk-lib/blob/638b6f48fe6643ec3e17ac6927d2f48fac96f663/lib/Dialect/Verif/IR/Ops.cpp#L1109-L1181).

One thing I found confusing is the use of `verif.require_compute` inside `verif.invariant`. `require_compute` still implements the precondition interface, and the existing contract verifier collects nested `PreconditionOpInterface` ops by walking the contract. So unless this is intentional, an invariant predicate written this way is still part of the contract-precondition machinery rather than a distinct invariant predicate language. See [`RequireOpBase`](https://github.com/project-llzk/llzk-lib/blob/638b6f48fe6643ec3e17ac6927d2f48fac96f663/include/llzk/Dialect/Verif/IR/Ops.td#L53-L67), [`PreconditionOpInterface`](https://github.com/project-llzk/llzk-lib/blob/638b6f48fe6643ec3e17ac6927d2f48fac96f663/include/llzk/Dialect/Verif/IR/OpInterfaces.td#L36-L42), the [contract verifier walk](https://github.com/project-llzk/llzk-lib/blob/638b6f48fe6643ec3e17ac6927d2f48fac96f663/lib/Dialect/Verif/IR/Ops.cpp#L644-L681), and the current [invariant test shape](https://github.com/project-llzk/llzk-lib/blob/638b6f48fe6643ec3e17ac6927d2f48fac96f663/test/Dialect/Verif/invariants_pass.llzk#L21-L26).

I also had questions about the structure of the new ops. Most of the invariant-related ops use ancestor checks rather than tighter placement rules. That makes it hard to tell whether nested invariants, nested steps, ranking ops inside steps, or broad `verif.old` usage are meant to be valid. If those are valid, it would help to spell out the semantics. If not, it seems worth rejecting them in the verifier. See the [invariant inner op definitions](https://github.com/project-llzk/llzk-lib/blob/638b6f48fe6643ec3e17ac6927d2f48fac96f663/include/llzk/Dialect/Verif/IR/Ops.td#L530-L612).

The loop target binding also seems worth clarifying. The syntax looks symbol-like, for example `for @loopA`, but the implementation matches a string against discardable `loop_label` attributes found by walking the target. That could be surprising because it reads like a scoped symbol reference, but does not seem to have symbol semantics. See the [`loop_label` lookup](https://github.com/project-llzk/llzk-lib/blob/638b6f48fe6643ec3e17ac6927d2f48fac96f663/include/llzk/Dialect/Verif/IR/OpInterfaces.td#L52-L92) and the [parser/printer](https://github.com/project-llzk/llzk-lib/blob/638b6f48fe6643ec3e17ac6927d2f48fac96f663/lib/Dialect/Verif/IR/Ops.cpp#L1183-L1247).

Some tests that would make the intended behavior clearer: loops with init/iter args, duplicate labels in different scopes, nested invariant/step cases, misplaced inner ops, `old` on constants or step-local values, and an explicit test showing whether `require_compute` inside an invariant is intended to be valid.

Overall I like the direction. I think this would be easier to build on if the invariant sublanguage had a more precise contract before more passes start depending on it.

include/llzk/Dialect/Verif/IR/OpInterfaces.td
0xddom: I can't find any reference to that name neither in the code nor in MLIR's. Where is that function/method defined?

iangneal: https://github.com/project-llzk/llzk-lib/blob/9303d0fc896b88dbeb04169acecfdc678270df07/include/llzk/Util/Walk.h#L36-L43

include/llzk/Dialect/Verif/IR/OpInterfaces.h
```diff
@@ -0,0 +1,23 @@
+//===-- OpInterfaces.h ------------------------------------------*- C++ -*-===//
+//
+// Part of the LLZK Project, under the Apache License v2.0.
+// See LICENSE.txt for license information.
+// Copyright 2025 Veridise Inc.
```
iangneal:
```suggestion
// Copyright 2026 Project LLZK
```

include/llzk/Dialect/Verif/IR/Ops.td
```diff
@@ -481,4 +491,134 @@ def IncludeOp
...
+// InvariantOp
+//===------------------------------------------------------------------===//
+
+def InvariantOp
+    : VerifDialectOp<"invariant", [HasAncestor<"::llzk::verif::ContractOp">,
+                                   NoTerminator, SingleBlock]> {
+
+  let summary = "loop invariant definition operation";
+  let description = [{
+    Defines an invariant for a loop inside the target of a contract.
```
iangneal, line 504: Maybe have some examples of the construction and clarify semantics here.

include/llzk/Dialect/Verif/IR/Ops.td
```diff
@@ -481,4 +491,134 @@ def IncludeOp
...
+  let builders = [OpBuilder<(ins "::mlir::StringRef":$loop_name,
+      CArg<"::llvm::ArrayRef<::mlir::Type>", "{}">:$loop_arg_types,
+      CArg<"::llvm::ArrayRef<::mlir::Location>", "{}">:$loop_arg_locs)>];
+
+  let hasCustomAssemblyFormat = 1;
+  let hasVerifier = 1;
+
+  let extraClassDeclaration = [{
+    /// Returns the contract operation that contains this invariant.
+    ::llzk::verif::ContractOp getParentContract();
```
iangneal, line 559: Would be nice to have a method for grabbing the targeted loop.

test/Dialect/Verif/invariants_pass.llzk
```diff
@@ -0,0 +1,248 @@
...
+    }
+    function.def @constrain(%self: !struct.type<@Top>)  {
+      function.return
+    }
+  }
+
+  verif.contract @Foo for @Top (%self: !struct.type<@Top>) {
+    verif.invariant for @loopA() {
+      %true = arith.constant  true
+      verif.require_compute %true
```
iangneal, line 24: I think I'm good with having precondition operations in `verif.invariant`s because it allows you to enforce preconditions over input arrays using the loop iteration constructs of the original target.

iangneal (changes requested):
Couple of minor things but otherwise LGTM

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -1255,4 +1226,28 @@ ContractOp InvariantOp::getParentContract() {
...
+  SmallVector<InvariantTargetOpInterface> matches;
+  for (auto invariantTarget : target->get().getLoops()) {
+    auto targetLabel = invariantTarget.getLabel();
+    if (succeeded(targetLabel) && *targetLabel == getLoopName()) {
+      matches.push_back(invariantTarget);
+    }
+  }
+
+  if (matches.size() == 0) {
+    return emitError() << "no invariant target with label \"" << getLoopName()
```
iangneal: You've got a lit test failure I believe because you're using `emitError` here but the lit test has `emitOpError` format.

iangneal (approved):
LGTM other than test failures.

include/llzk/Dialect/Verif/IR/Ops.td
```diff
@@ -481,4 +491,174 @@ def IncludeOp
...
+          %true = arith.constant  true
+          verif.require_compute %true
+        }
+      }
+    }
+    ```
+  }];
+
+  let arguments = (ins StrAttr:$loop_name, TypeArrayAttr:$loop_arg_types);
+  let results = (outs);
```
tim-hoffman: redundant

tim-hoffman: this appears several times in the file

include/llzk/Dialect/Verif/IR/Ops.td
```diff
@@ -481,4 +491,174 @@ def IncludeOp
...
+  let description = [{
+    Defines a predicate that must be satisfied between iterations of the loop.
+    
+    The predicate can access the value in the previous iteration by using the `verif.old` operation.
+
+    Declares the `MemoryEffectsOpInterface`, which, like the `cf.assert` (MLIR `cf` dialect)
+    and `bool.assert` (LLZK `bool` dialect) ops, adds a MemWrite affect to model program termination.
+  }];
+
+  let arguments = (ins);
```
tim-hoffman: redundant

lib/Dialect/Verif/IR/Ops.cpp
```diff
@@ -699,6 +704,13 @@ FailureOr<SymbolLookupResult<FuncDefOp>> ContractOp::getFuncTarget(SymbolTableCo
   );
 }
 
+FailureOr<SymbolLookupResult<ContractTargetOpInterface>>
+ContractOp::getTargetOp(SymbolTableCollection &tables) {
+  return lookupTopLevelSymbol<ContractTargetOpInterface>(
+      tables, getTarget(), getParentOfType<ModuleOp>(getOperation()), /*reportMissing*/ false
```
tim-hoffman, line 710: are you sure you shouldn't report if the symbol is missing? This is used in `InvariantOp::getTarget()` and other cases there emit an error but this one just silently returns failure. Is there a test that covers the `return failure();` in `InvariantOp::getTarget()`?

lib/Dialect/Verif/IR/Ops.cpp
0xddom: Oh, I know what went wrong. Such a dumb mistake.

lib/Dialect/Verif/IR/Ops.cpp
0xddom: Not that I could find. If I recall correctly I decided to set that one to false because the op would already fail validation if the target didn't exists and I wanted to avoid potentially duplicating the error message.

iangneal (approved):
LGTM

PR #545, opened 2026-06-14
Reject dynamically scoped scalar polynomial auxiliaries

### LINKED ISSUE #542: Poly lowering materializes loop-local expressions as one component-scope auxiliary
### Summary

`llzk-poly-lowering-pass` can materialize a loop-local high-degree expression as one component-scope auxiliary member. If the expression depends on an `scf.for` block argument, the lowered constraints can require that single auxiliary to equal different per-iteration values.

### Reproduction

Tested on current `main`:

```text
aced31cd0609fc26eaf0dbb0b608f7043c8d93b5
```

CI repro, from a clean fork branch whose diff from upstream `main` contains only the repro workflow and fixture:

```text
https://github.com/Kuhai9801/llzk-lib/actions/runs/27471141629
```

The run records:

```text
head=54c25ce9dd9f27a76eff441d9ae9624477b5c47a
upstream_main=aced31cd0609fc26eaf0dbb0b608f7043c8d93b5
verdict=lowered_with_loop_scoped_aux_artifact
```

Input:

```mlir
module attributes {llzk.lang = "llzk"} {
  struct.def @LoopAuxScope {
    struct.member @out : !felt.type {llzk.pub}

    function.def @compute(%a: !felt.type) -> !struct.type<@LoopAuxScope> {
      %self = struct.new : !struct.type<@LoopAuxScope>
      struct.writem %self[@out] = %a : !struct.type<@LoopAuxScope>, !felt.type
      function.return %self : !struct.type<@LoopAuxScope>
    }

    function.def @constrain(%self: !struct.type<@LoopAuxScope>, %a: !felt.type) {
      %lb = arith.constant 0 : index
      %ub = arith.constant 2 : index
      %step = arith.constant 1 : index
      %one = felt.const 1 : !felt.type

      %res = scf.for %iv = %lb to %ub step %step
          iter_args(%acc = %a) -> !felt.type {
        %sq = felt.mul %acc, %acc : !felt.type, !felt.type
        %cube = felt.mul %sq, %acc : !felt.type, !felt.type
        constrain.eq %cube, %cube : !felt.type
        %next = felt.add %acc, %one : !felt.type, !felt.type
        scf.yield %next : !felt.type
      }

      function.return
    }
  }
}
```

Run:

```sh
llzk-opt loop_aux_scope.llzk --llzk-poly-lowering-pass="max-degree=2"
```

### Actual behavior

The pass succeeds and emits one auxiliary member for the component:

```mlir
struct.member @__llzk_poly_lowering_pass_aux_member_0 : !felt.type
```

`compute` writes it once:

```mlir
%0 = felt.mul %arg0, %arg0 : !felt.type, !felt.type
struct.writem %self[@__llzk_poly_lowering_pass_aux_member_0] = %0
```

`constrain` then reads that same member inside the loop:

```mlir
%1 = felt.mul %arg3, %arg3 : !felt.type, !felt.type
%2 = struct.readm %arg0[@__llzk_poly_lowering_pass_aux_member_0]
constrain.eq %2, %1 : !felt.type, !felt.type
```

For `%a = 0`, the source constraint is tautological, but the lowered form requires the same auxiliary to equal `0 * 0` in the first iteration and `1 * 1` in the second.

### Expected behavior

The lowering should preserve the source relation. Either introduce region-aware/per-iteration auxiliary values, or reject nested-region block-argument dependent expressions with a diagnostic.

### Related

Related to #491 / #501, but distinct: those cover nested member-read reconstruction. This issue is about loop-local expression scope and nested-region block arguments.

1sgtpepper:
Fixes #542.

## Summary

Make the straight-line `@constrain` precondition explicit for polynomial and R1CS lowering.

The lowering passes materialize auxiliary values at component scope and assume control flow has already been flattened or otherwise lowered away. When a `constrain` body is not straight-line, the passes now fail with a diagnostic instead of rewriting region-local expressions into component-lifetime auxiliaries.

## Changes

- Add a shared `@constrain` body check that rejects multi-block bodies and nested-region or successor-bearing operations before auxiliary materialization.
- Apply the check to `llzk-poly-lowering-pass` and `llzk-r1cs-lowering`.
- Document the already-flattened straight-line precondition in the pass descriptions and full lowering pipeline summaries.
- Add regression tests for multi-block, loop, branch, and loop-local struct constrain call argument cases in poly lowering, plus loop and branch cases in R1CS lowering.
- Add a changelog entry for the new non-straight-line rejection behavior.

## Testing

- `git diff --check`
- `git clang-format --diff upstream/main HEAD --binary clang-format`
- GitHub Actions on `6aab49a989ca686d40ba837bcbc5b49fed4fdee6`: all checks passing

### DIFF
--- a/backends/r1cs/include/r1cs/Transforms/TransformationPasses.td
+++ b/backends/r1cs/include/r1cs/Transforms/TransformationPasses.td
@@ -18,7 +18,8 @@ def R1CSLoweringPass : LLZKPass<"llzk-r1cs-lowering"> {
   let description = [{
     Transforms LLZK constraints into an equivalent set of R1CS constraints expressed in the r1cs dialect.
     The pass expects already-flattened, straight-line `constrain` bodies. Run `llzk-flatten` or
-    another control-flow lowering pass before this pass.
+    another control-flow lowering pass before this pass; nested-region, successor-bearing, or
+    multi-block constraints are rejected before R1CS normalization.
     This pass is best used on already-flattened input as part of the `-llzk-full-r1cs-lowering`
     pipeline which includes a degree lowering pass and clean up passes to ensure correctness and
     performance.
--- a/backends/r1cs/include/r1cs/Transforms/TransformationPasses.td
+++ b/backends/r1cs/include/r1cs/Transforms/TransformationPasses.td
@@ -17,8 +17,11 @@ def R1CSLoweringPass : LLZKPass<"llzk-r1cs-lowering"> {
                 "i.e a*b - c = 0";
   let description = [{
     Transforms LLZK constraints into an equivalent set of R1CS constraints expressed in the r1cs dialect.
-    This pass is best used as part of the `-llzk-full-r1cs-lowering` pipeline which includes
-    a degree lowering pass and clean up passes to ensure correctness and performance.
+    The pass expects already-flattened, straight-line `constrain` bodies. Run `llzk-flatten` or
+    another control-flow lowering pass before this pass.
+    This pass is best used on already-flattened input as part of the `-llzk-full-r1cs-lowering`
+    pipeline which includes a degree lowering pass and clean up passes to ensure correctness and
+    performance.
   }];
 }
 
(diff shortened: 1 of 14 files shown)


1sgtpepper:
@shankarapailoor this is ready for another look when you have time. I updated the lowering guard to enforce the already-flattened straight-line precondition before aux materialization, and added coverage for the multi-block path as well as the loop/branch cases from #542.

shankarapailoor:
@codex review

shankarapailoor (approved):
LGTM

PR #546, opened 2026-06-15
Add quantifier ops

0xddom:


include/llzk/Dialect/Bool/IR/Ops.td
```diff
@@ -168,4 +169,50 @@ def LLZK_CmpOp
...
+
+  let hasCustomAssemblyFormat = 1;
+  let hasVerifier = 1;
+}
+
+def ForAllOp : QuantifierOp<"forall", "all elements"> {}
+
+def ExistsOp : QuantifierOp<"exists", "at least one element"> {}
+
+def YieldOp : BoolDialectOp<"yield", [NotFieldNative, Terminator,
```
tim-hoffman: The other "yield" ops in LLZK dialects use `ReturnLike` (although I'm not sure if it's really necessary on these ops or not).

include/llzk/Dialect/Bool/IR/Ops.td
```diff
@@ -168,4 +169,50 @@ def LLZK_CmpOp
...
+                                      ParentOneOf<["ForAllOp", "ExistsOp"]>]> {
+  let summary = "boolean yield operation";
+  let description = [{
+    This terminator operation yields the boolean value of the predicate inside the body of 
+    `bool.forall` and `bool.exists` operations.
+  }];
+
+  let arguments = (ins I1:$value);
+  let results = (outs);
```
tim-hoffman: redundant
```suggestion
```

include/llzk/Dialect/Bool/IR/Ops.td
```diff
@@ -168,4 +169,50 @@ def LLZK_CmpOp
...
+    This operation applies a predicate to each element of the given sort. 
+    Returns true iff the predicate is true for }]#cond#[{ of the sort.
+
+    Only LLZK arrays are supported as the sort type at the moment.
+
+    Example:
+    ```
+    %c10 = felt.const  10
+    %0 = array.new %a, %b, %c : !array.type<3 x !felt.type>
+    %1 = bool.}]#mnemonic#[{ %e: !felt.type in %0 {
```
tim-hoffman: Specifying the element type seems redundant since you could always infer from array element type.

tim-hoffman: Also this example is missing the `: type($sort)` syntax

unittests/CAPI/Dialect/Boolean.cpp
```diff
@@ -11,14 +11,23 @@
...
 #include "llzk/Dialect/Bool/IR/Attrs.capi.test.cpp.inc"
 #include "llzk/Dialect/Bool/IR/Dialect.capi.test.cpp.inc"
 #include "llzk/Dialect/Bool/IR/Enums.capi.test.cpp.inc"
 #include "llzk/Dialect/Bool/IR/Ops.capi.test.cpp.inc"
+#include "llzk/Dialect/Bool/IR/Ops.h"
+
+#include <llvm/ADT/Sequence.h>
+#include <llvm/ADT/SmallVectorExtras.h>
+
+#include <memory>
```
tim-hoffman, line 30: ordering of the includes vs comments about the includes is off

unittests/CAPI/Dialect/Boolean.cpp
```diff
@@ -140,3 +149,73 @@ std::unique_ptr<XorBoolOpBuildFuncHelper> XorBoolOpBuildFuncHelper::get() {
...
+            .create<llzk::array::CreateArrayOp>(
+                location,
+                llzk::array::ArrayType::get(llzk::felt::FeltType::get(builder.getContext()), {10}),
+                consts
+            )
+            .getResult()
+    );
+  }
+
+  /// Fills the body of the quantifier op to ensure its properly constructed.
```
tim-hoffman: nit
```suggestion
  /// Fills the body of the quantifier op to ensure it's properly constructed.
```

unittests/CAPI/Dialect/Boolean.cpp
```diff
@@ -140,3 +149,73 @@ std::unique_ptr<XorBoolOpBuildFuncHelper> XorBoolOpBuildFuncHelper::get() {
...
+  void fillBody(mlir::Operation *op, mlir::OpBuilder &builder, mlir::Location location) {
+    mlir::OpBuilder::InsertionGuard giard(builder);
```
tim-hoffman: nit
```suggestion
    mlir::OpBuilder::InsertionGuard guard(builder);
```

tim-hoffman (commented):
I would also like to see some lit tests where the array type has more than one dimension and something like the `ws` array in `test/FrontendLang/Circom/circom_example_2.llzk`

lib/Dialect/Bool/IR/Ops.cpp
```diff
@@ -152,4 +156,129 @@ OpFoldResult CmpOp::fold(FoldAdaptor adaptor) {
...
+  SMLoc loc = parser.getCurrentLocation();
+  if (parser.parseRegion(
+          *body, {arg},
+          /*enableNameShadowing=*/false
+      )) {
+    return failure();
+  }
+
+  if (body->empty()) {
+    return parser.emitError(loc, "expected non-empty invariant body");
```
tim-hoffman: also, "invariant" here doesn't make sense

include/llzk/Dialect/Bool/IR/Ops.td
0xddom: The example came before the parser and I had to tweak the syntax I initially had in mind. Had to switch it due to how operands are parsed and resolved

include/llzk/Dialect/Bool/IR/Ops.td
0xddom: I went back to check what that trait does and is related to control flow and matters if the parent implements `RegionBranchOpInterface`. Will depend on whether we want to treat the quantifiers. If they are something abstract that is only meant for lowering to the smt dialect then it's fine not to add those traits to the ops. If we want to be able to treat it as a 'for' loop that reduces the outputs with either `and` or `or` then the traits may help MLIR reason about them as loops.

0xddom:
> I would also like to see some lit tests where the array type has more than one dimension and something like the `ws` array in `test/FrontendLang/Circom/circom_example_2.llzk`

Good point. And that leads me to a question for @iangneal. What do we want to do in `llzk-spec` when we have a quantifier over a domain with more than one dimension? 

We can iterate over all the elements of the array:

```math
\begin{prooftree}
\RightLabel{\(\)}
\AxiomC{}
\UnaryInfC{\(a: \mathtt{Array<}N,M \times \tau_0\mathtt{>} \)}
\RightLabel{\(\)}
\AxiomC{}
\UnaryInfC{\(l: \tau_0\)}
\RightLabel{\(\)}
\AxiomC{}
\UnaryInfC{\(e: \mathtt{Bool}\)}
\RightLabel{\(\)}
\TrinaryInfC{\(\forall l \in a.e : \mathtt{Bool}\)}
\end{prooftree}
```

Or iterate over the first dimension:

```math
\begin{prooftree}
\RightLabel{\(\)}
\AxiomC{}
\UnaryInfC{\(a: \mathtt{Array<}N,M \times \tau_0\mathtt{>} \)}
\RightLabel{\(\)}
\AxiomC{}
\UnaryInfC{\(l: \mathtt{Array<}M \times \tau_0\mathtt{>}\)}
\RightLabel{\(\)}
\AxiomC{}
\UnaryInfC{\(e: \mathtt{Bool}\)}
\RightLabel{\(\)}
\TrinaryInfC{\(\forall l \in a.e : \mathtt{Bool}\)}
\end{prooftree}
```

The latter feels more intuitive to me from the point of view of the high level spec language. Another option could be to allow flexibility on the argument's type, allowing any type that is a subset of the array and picking the array element type as the default.

unittests/CAPI/Dialect/Boolean.cpp
```diff
@@ -19,6 +22,12 @@
 #include "llzk/Dialect/Bool/IR/Dialect.capi.test.cpp.inc"
 #include "llzk/Dialect/Bool/IR/Enums.capi.test.cpp.inc"
 #include "llzk/Dialect/Bool/IR/Ops.capi.test.cpp.inc"
+#include "llzk/Dialect/Bool/IR/Ops.h"
```
tim-hoffman: nit
```suggestion
#include "llzk/Dialect/Array/IR/Ops.h"
#include "llzk/Dialect/Bool/IR/Attrs.h"
#include "llzk/Dialect/Bool/IR/Ops.h"

// Include necessary generated CAPI
#include "llzk/Dialect/Bool/IR/Enums.capi.cpp.inc"

// Include the auto-generated tests
#include "llzk/Dialect/Bool/IR/Attrs.capi.test.cpp.inc"
#include "llzk/Dialect/Bool/IR/Dialect.capi.test.cpp.inc"
#include "llzk/Dialect/Bool/IR/Enums.capi.test.cpp.inc"
#include "llzk/Dialect/Bool/IR/Ops.capi.test.cpp.inc"
```

include/llzk/Dialect/Bool/IR/Ops.td
iangneal: I think it's probably reasonable to add `ReturnLike`, assuming it doesn't cause any issues in the current implementation. I could imagine that we might want to convert quantifiers into scf loops at some point, so it's worth adding the trait now if it's painless.

test/Dialect/Bool/logic_ops_pass.llzk
```diff
@@ -84,3 +84,79 @@ module attributes {llzk.lang} {
 // CHECK-NEXT:      }
 // CHECK-NEXT:    }
 // CHECK-NEXT:  }
+// -----
+
+function.def @f_forall(%a: !array.type<5 x index>) -> i1 attributes {function.allow_non_native_field_ops} {
+  %c = bool.forall %e in %a: !array.type<5 x index> {
```
iangneal, line 90: Can you add a test with a multi-dimensional array and clarify this is legal in the op documentation?

iangneal: Oh I see there was already a comment about this.

iangneal:
> > I would also like to see some lit tests where the array type has more than one dimension and something like the `ws` array in `test/FrontendLang/Circom/circom_example_2.llzk`
> 
> Good point. And that leads me to a question for @iangneal. What do we want to do in `llzk-spec` when we have a quantifier over a domain with more than one dimension?
> 
> We can iterate over all the elements of the array:
> a : Array < N , M × τ 0 > l : τ 0 e : Bool ∀ l ∈ a . e : Bool
> 
> Or iterate over the first dimension:
> a : Array < N , M × τ 0 > l : Array < M × τ 0 > e : Bool ∀ l ∈ a . e : Bool
> 
> The latter feels more intuitive to me from the point of view of the high level spec language. Another option could be to allow flexibility on the argument's type, allowing any type that is a subset of the array and picking the array element type as the default.

Iterating over the first dimension is probably better and easier to implement for now. I think the "iterating over any subset" is a nice extension of that concept, but not necessary for now.

0xddom:
> I would also like to see some lit tests where the array type has more than one dimension and something like the `ws` array in `test/FrontendLang/Circom/circom_example_2.llzk`

What part of `ws` did you want to have tests for? The template parameter or the map operand?

tim-hoffman:
> > I would also like to see some lit tests where the array type has more than one dimension and something like the `ws` array in `test/FrontendLang/Circom/circom_example_2.llzk`
> 
> What part of `ws` did you want to have tests for? The template parameter or the map operand?

Oops, I should have explained that more. I was thinking about two things. First, the array size is a defined by a template parameter. Second, that the element type uses an affine map because it's not homogeneous across the array. I want to make sure we're considering these nontrivial cases.

0xddom:
> > > I would also like to see some lit tests where the array type has more than one dimension and something like the `ws` array in `test/FrontendLang/Circom/circom_example_2.llzk`
> > 
> > 
> > What part of `ws` did you want to have tests for? The template parameter or the map operand?
> 
> Oops, I should have explained that more. I was thinking about two things. First, the array size is a defined by a template parameter. Second, that the element type uses an affine map because it's not homogeneous across the array. I want to make sure we're considering these nontrivial cases.

I think when it comes to the op themselves those things shouldn't have too bing of an impact, but I can add them for catching regressions if they turn out to be hairy in the future, specially regarding converting them to other ops or flattening.

lib/Dialect/Bool/IR/Ops.cpp
```diff
@@ -282,3 +283,15 @@ ParseResult ExistsOp::parse(OpAsmParser &parser, OperationState &result) {
 void ExistsOp::print(OpAsmPrinter &p) { printQuantOp(p, *this); }
 
 } // namespace llzk::boolean
+
+/// Extracts the type used for a quantifier op block argument.
+///
+/// If the array has only one dimension, returns the element type.
+/// Otherwise, returns an array type with the first dimension removed.
+Type llzk::boolean::getQuantifierOpDomainIterType(array::ArrayType arr) {
```
tim-hoffman, line 291: you can put this in the namespace block above and drop `llzk::boolean::` prefix

tim-hoffman (approved):
LGTM; just one nit

lib/Dialect/Bool/IR/Ops.cpp
0xddom: I did it that way on purpose because the compiler sometimes acts in a weird way with functions defined inside a namespace block if the declaration changes. I read about in the LLVM style [guide](https://llvm.org/docs/CodingStandards.html#use-namespace-qualifiers-to-define-previously-declared-symbols)

PR #549, opened 2026-06-17
`verif` to `smt` conversion

iangneal:
- Adds pass for fixpoint scalarization
- Adds a pass for converting scalarized `verif` contracts to `smt`
- Adds a pass pipeline for scalarizing and converting `verif` to `smt`
- Generalizes some function and call specific conversions to accept contracts and includes, since contracts are function-like and includes are call-like


include/llzk/Transforms/LLZKTransformationPasses.td
```diff
@@ -154,4 +154,28 @@ def WhileToForPass : LLZKPass<"llzk-while-to-for"> {
...
+  let summary =
+      "Repeatedly scalarize array/pod aggregates until the module stabilizes";
+  let description = [{
+    Repeatedly runs the existing array-to-scalar and pod-to-scalar passes,
+    followed by canonicalization, until the module reaches a fixpoint. This is
+    intended as a reusable preprocessing step for pipelines that need
+    array/pod-free IR.
+  }];
+  let options = [];
+}
```
tim-hoffman, line 140: I've been thinking about this a bit and I think we can augment `array-to-scalar` to split arrays within a pod (just create N pod records, similar to how splitting an array-type struct member) so that a single pass of `array-to-scalar` then `pod-to-scalar` can remove all.

iangneal: Since that'll take a bit more effort, can we add that as a TODO and keep this solution for now?

tim-hoffman: sure

include/llzk/Transforms/LLZKConversionUtils.h
```diff
@@ -183,6 +184,28 @@ inline function::CallOp createCallPreservingInstantiationOperands(
...
+inline verif::IncludeOp createIncludePreservingInstantiationOperands(
+    mlir::Location loc, verif::IncludeOp oldInclude, llvm::ArrayRef<mlir::ValueRange> mapOperands,
+    mlir::ValueRange argOperands, mlir::ConversionPatternRewriter &rewriter
+) {
+  llvm::SmallVector<mlir::Attribute> templateParams;
+  if (mlir::ArrayAttr templateParamsAttr = oldInclude.getTemplateParamsAttr()) {
+    templateParams.append(templateParamsAttr.begin(), templateParamsAttr.end());
+  }
+
+  if (oldInclude.getMapOperands().empty()) {
```
tim-hoffman: Should this be checking the `mapOperands` parameter instead?

include/llzk/Transforms/LLZKTransformationPasses.td
```diff
@@ -154,4 +154,28 @@ def WhileToForPass : LLZKPass<"llzk-while-to-for"> {
...
+def AggregateScalarizationPass : LLZKPass<"llzk-aggregate-scalarization"> {
+  let summary =
+      "Repeatedly scalarize array/pod aggregates until the module stabilizes";
+  let description = [{
+    Repeatedly runs the existing array-to-scalar and pod-to-scalar passes,
+    followed by canonicalization, until the module reaches a fixpoint. This is
+    intended as a reusable preprocessing step for pipelines that need
+    array/pod-free IR.
+  }];
+  let options = [];
```
tim-hoffman:
```suggestion
```

include/llzk/Transforms/LLZKTransformationPasses.td
```diff
@@ -154,4 +154,28 @@ def WhileToForPass : LLZKPass<"llzk-while-to-for"> {
...
+def VerifToSmtPass : LLZKPass<"llzk-scalar-verif-to-smt"> {
+  let summary =
+      "Lower scalarized verif contracts and simple LLZK targets to SMT helpers";
+  let description = [{
+    Lowers verif contracts into SMT helper functions for target computation,
+    preconditions, postconditions, and include implications. This pass currently
+    expects array/pod-free. Use the `-llzk-verif-to-smt` pipeline to preprocess
+    aggregate types first.
+  }];
+  let options = [];
```
tim-hoffman:
```suggestion
```

tim-hoffman:
@codex review

lib/Dialect/Array/Transforms/ArrayToScalarPass.cpp
```diff
@@ -306,6 +309,89 @@ void processInputOperands(
...
+        }
+      } else {
+        ++i;
+      }
+    }
+  }
+
+public:
+  explicit SplitArrayInFunctionLikeOpImpl(FunctionLikeOp op) {
+    inputNameInfo = collectSplitFunctionNameInfo(op.getArgumentTypes(), [&](unsigned i) {
```
tim-hoffman:
```suggestion
    inputNameInfo = collectSplitFunctionNameInfo(op.getArgumentTypes(), [&op](unsigned i) {
```

lib/Dialect/Array/Transforms/ArrayToScalarPass.cpp
```diff
@@ -306,6 +309,89 @@ void processInputOperands(
...
+      return op.getArgNameAttr(i);
+    }, getSplitArrayIndexSuffixes);
+    if constexpr (supportsResultAttrs()) {
+      ArrayAttr resultAttrs = op.getAllResultAttrs();
+      resultNameInfo = collectSplitFunctionNameInfo(op.getResultTypes(), [&](unsigned i) {
```
tim-hoffman:
```suggestion
      resultNameInfo = collectSplitFunctionNameInfo(op.getResultTypes(), [&resultAttrs](unsigned i) {
```

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -283,11 +286,92 @@ inline static void baseTargetSetup(ConversionTarget &target) {
...
+        }
+      } else {
+        ++i;
+      }
+    }
+  }
+
+public:
+  explicit SplitPodInFunctionLikeOpImpl(FunctionLikeOp op) {
+    inputNameInfo = collectSplitFunctionNameInfo(op.getArgumentTypes(), [&](unsigned i) {
```
tim-hoffman:
```suggestion
    inputNameInfo = collectSplitFunctionNameInfo(op.getArgumentTypes(), [&op](unsigned i) {
```

lib/Transforms/LLZKAggregateScalarizationPass.cpp
```diff
@@ -0,0 +1,262 @@
...
+#include <mlir/IR/BuiltinOps.h>
+#include <mlir/IR/Visitors.h>
+#include <mlir/Pass/PassManager.h>
+#include <mlir/Transforms/Passes.h>
+
+#include <llvm/ADT/DenseSet.h>
+#include <llvm/ADT/SmallVector.h>
+
+namespace llzk {
+#define GEN_PASS_DECL_AGGREGATESCALARIZATIONPASS
```
tim-hoffman: The general "#define GEN_PASS_DECL" + include in `include/llzk/Transforms/LLZKTransformationPasses.h` subsumes this
```suggestion
```

lib/Transforms/LLZKAggregateScalarizationPass.cpp
```diff
@@ -0,0 +1,262 @@
...
+  WalkResult result = module.walk([&](Operation *op) -> WalkResult {
+    if (auto structDef = dyn_cast<StructDefOp>(op)) {
+      llvm::SmallVector<Type> memberTypes;
+      for (MemberDefOp member : structDef.getMemberDefs()) {
+        memberTypes.push_back(member.getType());
+      }
+      if (failed(collectAggregateProfileForTypes(memberTypes, structDef, tables, profile))) {
+        return WalkResult::interrupt();
+      }
+      return WalkResult::advance();
```
tim-hoffman: You can remove all the intermediate `return WalkResult::advance();` by chaining these `if` with `else` so they all fall through to the final `return WalkResult::advance();`

test/Conversions/verif_to_smt_scalar_rejects_aggregates.llzk
```diff
@@ -0,0 +1,16 @@
+// RUN: llzk-opt --llzk-scalar-verif-to-smt %s -verify-diagnostics
+
+module attributes {llzk.lang} {
+  function.def @id_array(%arg: !array.type<2 x !felt.type>) -> !array.type<2 x !felt.type> {
+    function.return %arg : !array.type<2 x !felt.type>
+  }
+
+  // expected-error@+1 {{llzk-scalar-verif-to-smt requires array/pod-free IR for contract signature; run -llzk-verif-to-smt instead}}
```
tim-hoffman: this seems a little cleaner to me
```suggestion
  // expected-error@+1 {{'llzk-scalar-verif-to-smt' pass requires array/pod-free IR for contract signature; run 'llzk-verif-to-smt' instead}}
```

tim-hoffman (commented):
I have a few files to go through still but here's my feedback so far

include/llzk/Transforms/LLZKTransformationPasses.td
tim-hoffman: Update, Raghav needed all pod types removed without also removing array types so [this PR](https://github.com/project-llzk/llzk-lib/pull/565) increases the capabilities of `pod-to-scalar` such that running it should be able to remove all `pod.type` even within an array so pipelines can just run it before `array-to-scalar`.

iangneal:
@codex review

iangneal:
@codex review

iangneal:
@codex review

include/llzk/Transforms/LLZKConversionUtils.h
```diff
@@ -201,7 +224,7 @@ class FunctionTypeConverter {
 public:
   virtual ~FunctionTypeConverter() = default;
 
-  void convert(function::FuncDefOp op, mlir::RewriterBase &rewriter) {
+  template <typename FunctionLikeOp> void convert(FunctionLikeOp op, mlir::RewriterBase &rewriter) {
```
0xddom, line 227: Change the docstring of the class now that it works on any function-like thing

include/llzk/Util/Walk.h
```diff
@@ -33,6 +33,19 @@ template <typename MatchType, typename R> inline static bool walkContains(R &roo
...
 
+/// Returns whether the MLIR walk rooted at `root` contains any operation whose
+/// concrete type is one of `MatchTypes`.
+///
+/// Traversal stops at the first matching operation.
+template <typename MatchType, typename NextMatchType, typename... MatchTypes, typename R>
+inline static bool walkContains(R &root) {
+  return root
+      .walk([](mlir::Operation *op) {
+    return llvm::isa<MatchType, NextMatchType, MatchTypes...>(op) ? mlir::WalkResult::interrupt()
```
0xddom, line 44: Why do you need `typename MatchType, typename NextMatchType`? That forces callers to check for at least 2 possibilities.

lib/Dialect/Array/Transforms/ArrayToScalarPass.cpp
```diff
@@ -306,6 +309,90 @@ void processInputOperands(
...
+class SplitArrayInFunctionLikeOpImpl : public FunctionTypeConverter {
+  SmallVector<size_t> originalInputIdxToSize, originalResultIdxToSize;
+  SplitFunctionNameInfo inputNameInfo;
+  SplitFunctionNameInfo resultNameInfo;
+
+  static constexpr bool supportsResultAttrs() {
+    return requires(FunctionLikeOp op, ArrayAttr attrs) {
+      op.getResAttrsAttr();
+      op.setResAttrsAttr(attrs);
+    };
```
0xddom, line 330: This pattern is used in `LLZKConversionUtils.h` as well so we could move the function there to share it.

lib/Dialect/Array/Transforms/ArrayToScalarPass.cpp
```diff
@@ -548,12 +580,38 @@ class SplitArrayInCallOp : public OpConversionPattern<CallOp> {
   void rewrite(CallOp op, OpAdaptor adaptor, ConversionPatternRewriter &rewriter) const override {
     // Create new CallOp with split results first so, then process its inputs to split types
     CallOp newCall = newCallOpWithSplitResults(op, adaptor, rewriter);
+    rewriter.setInsertionPoint(newCall);
```
0xddom, line 583: This is such that whatever is created next in `processInputOperands` is inserted before `newCall`, right? Don't we need to recover the previous insertion point after that? Otherwise it will continue inserting before the call unless the conversion engine saves the insertion point internally before calling a pattern's rewrite function.

lib/Transforms/LLZKAggregateScalarizationPass.cpp
```diff
@@ -0,0 +1,252 @@
...
+        return WalkResult::interrupt();
+      }
+    } else if (auto newPod = dyn_cast<NewPodOp>(op)) {
+      if (failed(collectAggregateProfileForTypes(
+              ArrayRef<Type> {newPod.getType()}, newPod, tables, profile
+          ))) {
+        return WalkResult::interrupt();
+      }
+    }
+    return WalkResult::advance();
```
0xddom, line 193: You can use `llvm::TypeSwitch` here instead of chaining if-then-elses

lib/Transforms/LLZKAggregateScalarizationPass.cpp
```diff
@@ -0,0 +1,252 @@
...
+  }
+  return profile;
+}
+
+class AggregateScalarizationPass
+    : public llzk::impl::AggregateScalarizationPassBase<AggregateScalarizationPass> {
+public:
+  void runOnOperation() override {
+    ModuleOp module = getOperation();
+    constexpr unsigned kMaxIterations = 8;
```
0xddom, line 207: If we are gonna have a hard iteration limit I'd make it a pass parameter. That way users can tweak it if they need more iterations.

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+///
+/// 4. `@smt_verif_<Contract>_{compute,constrain}_entry`, which prove each direct contract stage
+///    by checking that the negated condition is unsat before asserting the condition into the
+///    caller solver context.
+///
+/// The pass runs as a sequential module transformation. For each contract in source order it
+/// validates the contract and its direct target, creates any missing target helpers, materializes
+/// the contract-local helpers, creates any per-include wrappers, and finally emits the full entry
+/// helper. Once every contract has been lowered successfully, the original `verif.contract`
+/// operations are erased from the module.
```
0xddom, line 40: If the contracts don't need to see each other as they are transformed this pass could probably be implemented as a pass over `verif.contract`, enabling parallelism. If the `verif.include` ops can deterministically figure out what the transformed result of the callee `verif.contract` is going to look like and create the call to the function then they should be able to do it in parallel.

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+
+  auto &block = region.front();
+  if (!isa<scf::YieldOp>(block.getTerminator())) {
+    return ifOp.emitOpError() << "expects " << branchName
+                              << " region to terminate with scf.yield while lowering " << context
+                              << " to SMT";
+  }
+
+  return success();
+}
```
0xddom, line 111: If this check really necessary? The ODS of `scf.if` specifically says that this is the expected shape, so an `scf.if` op that passed verification always passes this check

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+    return failure();
+  }
+  if (!requireElseRegion && ifOp.getElseRegion().empty()) {
+    return success();
+  }
+  if (failed(validateSingleBlockRegion(ifOp, ifOp.getElseRegion(), "else", context))) {
+    return failure();
+  }
+  return success();
+}
```
0xddom, line 125: Same thing with this one, seems redundant unless they are being created by this pass.

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+
+  /// Symbol name of the direct postcondition helper.
+  std::string post;
+
+  /// Symbol name of the full contract entry helper.
+  std::string entry;
+
+  /// Symbol names of the per-include wrapper helpers in source order.
+  SmallVector<std::string> includeHelpers;
+};
```
0xddom, line 168: If these names come from symbol/string attributes in operations they can be `llvm::StringRef` instead of creating copies with `std::string` since the strings live in the context's arena.

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+      first = false;
+      os << piece;
+    }
+    return stem;
+  }
+
+  /// Lower one LLZK scalar type to the corresponding SMT-facing type.
+  Type lowerType(Type type) {
+    if (isa<FeltType>(type)) {
+      return smt::IntType::get(context);
```
0xddom, line 370: You mentioned something about the conversion to and from bitvecs was slowing down the solver. If we know the field the felt type is using we could use bitvecs to represent them, so I wonder what would the speed difference be between using `int` and bitvecs for every operation.

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+LoweringContext::getStructMembers(StructType type, Operation *origin) {
+  auto structDef = resolveStructDef(type, origin);
+  if (failed(structDef)) {
+    return failure();
+  }
+  SmallVector<MemberDefOp> members;
+  for (MemberDefOp member : structDef->getMemberDefs()) {
+    members.push_back(member);
+  }
+  return members;
```
0xddom, line 448:
```suggestion
  SmallVector<MemberDefOp> members(structDef->getMemberDefs());
  return members;
```

Alternatively, we could change `StructDefOp::getMemberDefs` to return `SmallVector<MemberDefOp>` instead of `std::vector<MemberDefOp>`.

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+    return failure();
+  }
+
+  auto loweredMemberIt = parentIt->second.find(memberRead.getMemberName());
+  if (loweredMemberIt == parentIt->second.end()) {
+    emitError(
+        aggregate.getLoc(), (Twine("missing lowered member @") + memberRead.getMemberName() +
+                             " while materializing nested struct mapping")
+                                .str()
+    );
```
0xddom, line 574: Can't this line be written as `emitError(aggregate.getLoc()) << ...`?

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+      lowered.push_back(builder.create<smt::IntConstantOp>(origin->getLoc(), intAttr).getResult());
+      continue;
+    }
+    if (auto feltAttr = dyn_cast<FeltConstAttr>(attr)) {
+      lowered.push_back(builder
+                            .create<smt::IntConstantOp>(
+                                origin->getLoc(), IntegerAttr::get(
+                                                      builder.getContext(),
+                                                      toAPSInt(toDynamicAPInt(feltAttr.getValue()))
+                                                  )
```
0xddom, line 657: This conversion to `IntegerAttr` could be a method in `FeltConstAttr`

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+      continue;
+    }
+    if (auto refAttr = dyn_cast<FlatSymbolRefAttr>(attr)) {
+      auto it = constParamMap.find(refAttr.getRootReference().strref());
+      if (it == constParamMap.end()) {
+        origin->emitError().append(
+            "missing template parameter binding for @", refAttr.getRootReference(),
+            " in verif-to-smt lowering"
+        );
+        return failure();
```
0xddom, line 669: nit
```suggestion
        return origin->emitError() 
            << "missing template parameter binding for @" 
            << refAttr.getRootReference() 
            << " in verif-to-smt lowering";
```

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+      }
+      lowered.push_back(it->second);
+      continue;
+    }
+
+    origin->emitError("unsupported non-type template parameter in verif-to-smt lowering");
+    return failure();
```
0xddom, line 676: nit
```suggestion
    return origin->emitError("unsupported non-type template parameter in verif-to-smt lowering");
```

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+  for (Type input : type.getInputs()) {
+    if (failed(ensureScalarTypeSupported(input, origin, description))) {
+      return failure();
+    }
+  }
+  for (Type result : type.getResults()) {
+    if (failed(ensureScalarTypeSupported(result, origin, description))) {
+      return failure();
+    }
+  }
```
0xddom, line 711:
```suggestion
  for (Type type : llvm::concat(type.getInputs(), type.getResults())) {
    if (failed(ensureScalarTypeSupported(type, origin, description))) {
      return failure();
    }
  }
```

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+    current = builder.create<smt::AndOp>(loc, current, condition).getResult();
+  }
+  return current;
+}
+
+/// Build element-wise equality conditions between two equally-sized value ranges.
+static SmallVector<Value>
+buildEqualityConditions(OpBuilder &builder, Location loc, ValueRange lhs, ValueRange rhs) {
+  SmallVector<Value> conditions;
+  for (auto [lhsValue, rhsValue] : llvm::zip(lhs, rhs)) {
```
0xddom, line 817: To ensure the same size precondition in the docstring.
```suggestion
  for (auto [lhsValue, rhsValue] : llvm::zip_equal(lhs, rhs)) {
```

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+
+/// Return `true` if `op` is nested under an `scf.if` before reaching `scope`.
+static bool hasIfAncestor(Operation *op, Operation *scope) {
+  for (Operation *ancestor = op->getParentOp(); ancestor && ancestor != scope;
+       ancestor = ancestor->getParentOp()) {
+    if (isa<scf::IfOp>(ancestor)) {
+      return true;
+    }
+  }
+  return false;
```
0xddom, line 842:
```suggestion
  return op->getParentOfType<scf::IfOp>();
```

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+      for (auto [orig, lowered] : llvm::zip(call.getResults(), loweredCall.getResults())) {
+        valueMap[orig] = lowered;
+      }
+      return loweredCall.getResult(cast<OpResult>(value).getResultNumber());
+    }
+
+    definingOp->emitError().append(
+        "unsupported expression in verif-to-smt lowering: ", definingOp->getName().getStringRef()
+    );
+    return failure();
```
0xddom, line 1291: This whole mess is what patterns are for...

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+    }
+
+    Value constant = builder
+                         .create<smt::IntConstantOp>(
+                             loc, IntegerAttr::get(builder.getContext(), toAPSInt(value))
+                         )
+                         .getResult();
+    intConstantCache.try_emplace(std::move(key), constant);
+    return constant;
+  }
```
0xddom, line 1323: `llvm::DynamicAPInt` has a `hash_code` friend function so iirc it should be able to be used as a hash table key instead of creating and allocating a string representation of the integer **every time** just to save creating a duplicate int constant op.

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+    llvm::raw_svector_ostream(repr) << value;
+    return std::string(repr);
+  }
+
+  /// Convert a felt constant attribute into the equivalent SMT integer literal.
+  IntegerAttr toIntAttr(FeltConstantOp op) {
+    return IntegerAttr::get(
+        builder.getContext(), toAPSInt(toDynamicAPInt(op.getValue().getValue()))
+    );
+  }
```
0xddom, line 1307: The docstring and the function argument are at odds with each other. Is the function meant to take an op or an attribute?

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+
+  /// Lower signed integer division over field elements.
+  Value emitSignedIntDivValue(Location loc, Value lhs, Value rhs, FeltType type) {
+    return emitSignedDivOrRem(loc, lhs, rhs, type, /*isDiv=*/true);
+  }
+
+  /// Lower signed remainder over field elements.
+  Value emitSignedModValue(Location loc, Value lhs, Value rhs, FeltType type) {
+    return emitSignedDivOrRem(loc, lhs, rhs, type, /*isDiv=*/false);
+  }
```
0xddom, line 1445: `emitSignedDivOrRem` doesn't seem to be used anywhere else so this looks over-complicated for no reason

lib/Transforms/LLZKVerifToSmtPass.cpp
```diff
@@ -0,0 +1,2516 @@
...
+  return guarded;
+}
+
+/// Reject struct constrain bodies that cannot be summarized as one boolean formula.
+static LogicalResult checkStructConstrainBodyIsBooleanSummarizable(FuncDefOp constrainFunc) {
+  bool failed = false;
+  constrainFunc.walk([&](Operation *op) {
+    if (op == constrainFunc.getOperation() || isa<ReturnOp, scf::YieldOp>(op) ||
+        isa<constrain::EmitEqualityOp>(op) || isa<CallOp>(op)) {
```
0xddom, line 1615:
```suggestion
    if (op == constrainFunc.getOperation() || isa<ReturnOp, scf::YieldOp, constrain::EmitEqualityOp, CallOp>(op)) {
```

test/Conversions/verif_to_smt_regressions.llzk
0xddom, line 1: The organization in this file is confusing me

lib/Transforms/LLZKVerifToSmtPass.cpp
0xddom, line 1: Are we planning to support lowering the loop invariant ops in this PR?

0xddom (changes requested):
The verif to smt pass needs some work. A cleaner approach is to let the different dialects define how to lower to smt, similar to how lowering to LLVM works in upstream MLIR. The pass' name hints that is going to lower only operations from the verif dialect, but it lowers everything that is meant to be lowered to smt. The pass may probably be better renamed to `-llzk-to-smt`.

The current lowering logic is pretty fragile. If we add new ops (like `bool.forall` coming soon) then we need to change this pass as well to support it. If the implementation for converting from LLZK to SMT is defined by the ops or the dialect, then we only need to change the dialect or the op, a change probably closer to where the other changes that break the pass are happening.

The code as it stands right now is a good prototype for getting the demo working for yesterday but that's about it.

iangneal:
> The verif to smt pass needs some work. A cleaner approach is to let the different dialects define how to lower to smt, similar to how lowering to LLVM works in upstream MLIR. The pass' name hints that is going to lower only operations from the verif dialect, but it lowers everything that is meant to be lowered to smt. The pass may probably be better renamed to `-llzk-to-smt`.

Problem is we already have a lowering pass called that, but it's specifically for @product aligned programs. We can rename that pass, perhaps, but that would require an LLEQ change. @raghav198 

Otherwise I agree.

> The code as it stands right now is a good prototype for getting the demo working for yesterday but that's about it.

Correct. I think there are some PRs that actually need to be split off from this one and merged separately or removed. For example, there's a scalarization workaround that I believe needs to be reverted once https://github.com/project-llzk/llzk-lib/pull/565 is merged.

I'm just going to mark this as a draft until I'm done restructuring this, but I haven't had a chance to yet.

iangneal:
Handing this off to @raghav198.

PR #555, opened 2026-06-17
Replace copied SparseAnalysis with MLIR adapter

### LINKED ISSUE #308: Investigate alternatives to ported SparseAnalysis
In this PR: https://github.com/project-llzk/llzk-lib/pull/307, we removed the ported DenseAnalysis since we no longer needed a custom implementation.

The SparseAnalysis was ported for a similar reason, but also has changes to the (non-virtual) `visitOperation` function so that operations without results (constraints, member writes, etc) are visited.

https://github.com/project-llzk/llzk-lib/blob/67030b1fd8900c252bede1c9b68dd81d99dae264/lib/Analysis/SparseAnalysis.cpp#L102-L105

We can likely implement this functionality in a different way tnat allows us to remove this port.

1sgtpepper:
Refs #308.

This removes LLZK's copied sparse data-flow implementation and replaces it with a small adapter over upstream MLIR sparse analysis.

The adapter preserves LLZK's required handling for live zero-result operations, including operand subscriptions, external zero-result calls, and interprocedural zero-result call behavior.

Added focused unit coverage for zero-result operation revisits, program-order initialization, liveness, region branches, and call handling.

Validation:
- Fork CI: https://github.com/Kuhai9801/llzk-lib/pull/3
- `git diff --check`
- Formatting, docs, Nix debug/release, GCC tests, and Clang tests passed


lib/Analysis/SparseAnalysis.cpp
```diff
@@ -1,337 +1,157 @@
-//===- SparseAnalysis.cpp - Sparse data-flow analysis -----------*- C++ -*-===//
+//===- SparseAnalysis.cpp - LLZK sparse data-flow adapter -----------------===//
 //
 // Part of the LLZK Project, under the Apache License v2.0.
 // See LICENSE.txt for license information.
-// Copyright 2025 Veridise Inc.
+// Copyright 2026 Project LLZK
```
iangneal, line 6: Nit: don't erase copyright history.
```suggestion
// Copyright 2025 Veridise Inc.
// Copyright 2026 Project LLZK
```

unittests/Analysis/SparseAnalysisTests.cpp
```diff
@@ -0,0 +1,530 @@
...
+  using Base = llzk::dataflow::SparseForwardDataFlowAnalysis<TestSparseLattice>;
+
+public:
+  using Base::Base;
+
+  LogicalResult visitOperation(
+      Operation *op, ArrayRef<const TestSparseLattice *> /*operands*/,
+      ArrayRef<TestSparseLattice *> /*results*/
+  ) override {
+    if (op->getName().getStringRef() == kFunctionCallName) {
```
iangneal: Nit: instead of using the op name, just use "isa"
```suggestion
    if (isa<llzk::function::CallOp>(op)) {
```

unittests/Analysis/SparseAnalysisTests.cpp
```diff
@@ -0,0 +1,530 @@
...
+    StringRef opName = op->getName().getStringRef();
+
+    if (opName == kDelayedProducerName) {
+      // Make the producer depend on the trigger. The first producer visit
+      // leaves the result lattice at its optimistic value. After the sink sets
+      // the trigger, the producer is revisited and changes its result lattice.
+      TriggerState *trigger = getTriggerState();
+      addDependency(trigger, getProgramPointAfter(op));
+      if (!results.empty() && trigger->isReady()) {
+        propagateIfChanged(results.front(), results.front()->setValue(42));
```
iangneal: Nit: use named constants for these test values instead of magic numbers

lib/Analysis/SparseAnalysis.cpp
```diff
@@ -1,337 +1,157 @@
...
+#include <mlir/Support/LogicalResult.h>
 
-#include <llvm/ADT/STLExtras.h>
+#include <llvm/ADT/SmallVector.h>
 #include <llvm/Support/Casting.h>
 
-#include <cassert>
-#include <optional>
-
-using namespace mlir;
```
iangneal, line 41: Nit: have `using namespace mlir;` reduces verbosity. We do this pretty consistently throughout the codebase, so I'd keep the namespace usage (you may have to specific `::mlir::dataflow::...` for a few things but I think we'd prefer that).

iangneal (changes requested):
Thanks for the contribution! Overall this LGTM, I just have a few nits.

iangneal (approved):
LGTM---thank you for your contribution!

PR #560, opened 2026-06-17
Reject non-straight-line poly compute bodies

### LINKED ISSUE #554: Poly lowering inserts compute aux write using value from non-dominating block
### Summary

For a `compute` function with multiple blocks, `-llzk-poly-lowering-pass --verify-each` can insert an aux write using an operand that does not dominate the insertion point.

### Reproduction

Fork CI run:
https://github.com/Kuhai9801/llzk-lib/actions/runs/27666093229

Branch:
`ci/poly-repro-compute-multiblock`

Captured verifier error:

```text
operand #0 does not dominate this use
```

### Expected

Aux writes should be inserted only in regions where all operands dominate, or multiblock compute functions should be rejected before rewriting.

### Actual

The pass emits dominance-invalid IR for a multiblock compute function.

1sgtpepper:
Fixes #554.

## Summary
Polynomial lowering now rejects non-straight-line compute bodies before inserting aux writes at the compute terminator.

## Changes
- Generalize the straight-line body check for named function bodies.
- Apply the check to compute before aux materialization.
- Document the compute/constrain straight-line precondition.
- Add a multiblock compute regression and an unreleased changelog entry.

## Testing
- Fork CI core jobs on `Kuhai9801/llzk-lib@f5ce0f55802af11fea3f0b79d78be881ea662ccd` passed: `build-docs`, `run-clang-format`, Clang/GCC, `debug`, `release`, and `nix-check-result`.


lib/Transforms/LLZKLoweringUtils.cpp
```diff
@@ -99,16 +99,15 @@ LogicalResult checkForAuxMemberConflicts(StructDefOp structDef, StringRef prefix
...
 }
 
-LogicalResult checkConstrainBodyIsStraightLine(FuncDefOp constrainFunc, StringRef passName) {
-  auto emitStraightLineError = [passName](Operation *op) {
-    op->emitError() << passName
-                    << " expects a straight-line constrain body; run `llzk-flatten` or another "
-                       "control-flow lowering pass first";
+LogicalResult checkFuncBodyIsStraightLine(FuncDefOp func, StringRef passName, StringRef funcName) {
+  auto emitStraightLineError = [passName, funcName](Operation *op) {
+    op->emitError() << passName << " expects a straight-line " << funcName
```
iangneal, line 158: FuncDefOp has methods for `isStructCompute()`/`isStructConstrain()`/`isStructCompute()` that you could use here rather than having the caller pass in a `funcName`. This way you could remove the specialization for `checkConstrainBodyIsStriaghtLine` and have a single `checkFuncBodyIsStraightLine`

iangneal (changes requested):
Overall LGTM, just one comment:

iangneal (approved):
LGTM, thanks for the contribution!

1sgtpepper:
Thanks for the review!

PR #564, opened 2026-06-18
use `StringAttr` instead of `FlatSymbolRefAttr` in pod access ops

tim-hoffman:
use `StringAttr` instead of `FlatSymbolRefAttr` for record name in pod read/write ops


iangneal (approved):
Neato

PR #565, opened 2026-06-18
Update `llzk-pod-to-scalar` to split pods within arrays to multiple arrays

tim-hoffman:


tim-hoffman:
@codex review

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,6 +221,107 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
+    if (PodType nestedPodTy = dyn_cast<PodType>(record.getType())) {
+      forEachPodLeaf(nestedPodTy, recordChain, callback);
+    } else {
+      callback(RecordChain(recordChain), record.getType());
+    }
+    recordChain.pop_back();
+  }
+}
+
+/// If the given ArrayType has a POD element type, return it.
```
iangneal:
```suggestion
/// If the given ArrayType has a POD element type, return the ArrayType.
```
"it" made me think you were referring to the POD element type

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,6 +221,107 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
+inline static ArrayType splittablePodArray(ArrayType at) {
+  return isa<PodType>(at.getElementType()) ? at : nullptr;
+}
+
+/// If the given Type is an ArrayType with a POD element type, return it.
```
iangneal:
```suggestion
/// If the given Type is an ArrayType with a POD element type, return the parent ArrayType.
```
Same as above.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,6 +221,107 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
 
+/// Visit each non-POD leaf record in `podTy`, providing its record-name chain and leaf type.
+template <typename Fn>
+static void forEachPodLeaf(PodType podTy, SmallVectorImpl<StringAttr> &recordChain, Fn &&callback) {
+  for (RecordAttr record : podTy.getRecords()) {
+    recordChain.push_back(record.getName());
+    if (PodType nestedPodTy = dyn_cast<PodType>(record.getType())) {
+      forEachPodLeaf(nestedPodTy, recordChain, callback);
+    } else {
+      callback(RecordChain(recordChain), record.getType());
```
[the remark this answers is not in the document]
iangneal: Yeah I think we should have a test case with this example.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,6 +221,107 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
+/// If `t` is an array with POD element type, append one parallel array type for each POD leaf.
+static size_t splitPodArrayTypeTo(
+    Type t, SmallVectorImpl<Type> &collect, SmallVector<RecordChain> *splitIds = nullptr
+) {
+  if (ArrayType at = splittablePodArray(t)) {
+    auto podTy = llvm::cast<PodType>(at.getElementType());
+    SmallVector<StringAttr> recordChain;
+    size_t originalSize = collect.size();
+    forEachPodLeaf(podTy, recordChain, [&](RecordChain id, Type leafType) {
+      collect.push_back(at.cloneWith(leafType));
```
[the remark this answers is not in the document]
iangneal: I don't think we have this test case either, we should add it.

iangneal (dismissed):

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,203 +679,3209 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
-using LocalMemberReplacementMap = DenseMap<RecordChain, MemberInfo>;
-/// struct -> original pod-type member name -> LocalMemberReplacementMap
-using MemberReplacementMap = DenseMap<StructDefOp, DenseMap<StringAttr, LocalMemberReplacementMap>>;
+/// Materialize any missing synthetic backing components for one deferred fresh `pod.read`.
+static DeferredPodArrayBacking &materializeDeferredPodArrayBacking(
+    ReadPodOp readOp, ArrayType arrTy, ArrayRef<Type> splitTypes,
+    DeferredPodArrayBackingMap &deferredPodArrays, Location loc, OpBuilder &bldr,
+    bool requireLeafArrays, bool requireShapeCarrier
+) {
+  auto [it, inserted] = deferredPodArrays.try_emplace(readOp.getResult());
```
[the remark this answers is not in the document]
tim-hoffman: Two reads do not give aliases to shared data. Read is essentially a copy.

tim-hoffman:
@codex review

test/Transforms/PodToScalar/constraints_and_whole_pod_uses.llzk
```diff
@@ -0,0 +1,653 @@
...
+
+    function.def @constrain(%self: !struct.type<@Box>, %old: index, %new: index) {
+      function.return
+    }
+  }
+}
+// CHECK-LABEL: module attributes {llzk.lang} {
+// CHECK-NEXT:    struct.def @Box {
+// CHECK-NEXT:      struct.member @p_x : index
+// CHECK-NEXT:      struct.member @out : index
```
0xddom, line 298: Do we have a test for collisions when expanding the name of the member? Imagine the case where `@Box` already has a member named `@p_x`.

test/Transforms/PodToScalar/constraints_and_whole_pod_uses.llzk
```diff
@@ -0,0 +1,653 @@
...
+  }
+}
+// CHECK-LABEL: module attributes {llzk.lang} {
+// CHECK-NEXT:    poly.template @ContainArrayPodAndTypeVar {
+// CHECK-NEXT:      poly.param @E : !poly.tvar<@E>
+// CHECK-NEXT:      function.def @constrain(
+// CHECK-SAME:         %[[ARR_X:[0-9a-zA-Z_\.]+]]: !array.type<2 x index>,
+// CHECK-SAME:         %[[ARR_Y:[0-9a-zA-Z_\.]+]]: !array.type<2 x !felt.type>,
+// CHECK-SAME:         %[[ELEM:[0-9a-zA-Z_\.]+]]: !poly.tvar<@E>) attributes {function.allow_constraint} {
+// CHECK-NEXT:        %[[ELEM_LEAVES:[0-9a-zA-Z_\.]+]]:2 = builtin.unrealized_conversion_cast %[[ELEM]] : !poly.tvar<@E> to index, !felt.type
```
0xddom: I'm not sure if this cast can be easily reconciled without special handling since it goes 1:2 in terms of values. What happens if we feed this IR to the flattening pass?

test/Transforms/PodToScalar/constraints_and_whole_pod_uses.llzk
```diff
@@ -0,0 +1,653 @@
...
+  function.def @nested_eq_before_later_write(
+      %cond: i1, %lhs: !Pair, %rhs: !Pair, %new: index
+  ) attributes {function.allow_constraint} {
+    scf.if %cond {
+      constrain.eq %lhs, %rhs : !Pair
+    }
+    pod.write %lhs[@y] = %new : !Pair, index
+    function.return
+  }
+}
```
0xddom, line 415: A more interesting test would be to have the write inside the conditional.

test/Transforms/PodToScalar/constraints_and_whole_pod_uses.llzk
```diff
@@ -0,0 +1,653 @@
...
+    }
+  }
+}
+// CHECK-LABEL: module attributes {llzk.lang} {
+// CHECK-NEXT:    poly.template @ContainGenericArrayAndPodSubarray {
+// CHECK-NEXT:      poly.param @T : !poly.tvar<@T>
+// CHECK-NEXT:      function.def @constrain(%[[VAL_0:[0-9a-zA-Z_\.]+]]: !array.type<2,? x !poly.tvar<@T>>, %[[VAL_1:[0-9a-zA-Z_\.]+]]: !array.type<? x index>, %[[VAL_2:[0-9a-zA-Z_\.]+]]: !array.type<? x !felt.type>, %[[VAL_3:[0-9a-zA-Z_\.]+]]: !array.type<? x none>) attributes {function.allow_constraint} {
+// CHECK-NEXT:        %[[VAL_4:[0-9a-zA-Z_\.]+]] = poly.unifiable_cast %[[VAL_0]] : (!array.type<2,? x !poly.tvar<@T>>) -> !array.type<2,? x index>
+// CHECK-NEXT:        %[[VAL_5:[0-9a-zA-Z_\.]+]] = poly.unifiable_cast %[[VAL_0]] : (!array.type<2,? x !poly.tvar<@T>>) -> !array.type<2,? x !felt.type>
+// CHECK-NEXT:        %[[VAL_6:[0-9a-zA-Z_\.]+]] = poly.unifiable_cast %[[VAL_0]] : (!array.type<2,? x !poly.tvar<@T>>) -> !array.type<2,? x none>
```
0xddom: This cast feel wrong when compared to the others

test/Transforms/PodToScalar/function_calls_with_pod.llzk
```diff
@@ -159,11 +174,166 @@ module attributes {llzk.lang} {
...
+!PodArray = !array.type<2 x !Pair>
+module attributes {llzk.lang} {
+  function.def @named_array_result(%arg: !PodArray) -> (!PodArray {function.res_name = "out"}) {
+    function.return %arg : !PodArray
+  }
+}
+// CHECK-LABEL: module attributes {llzk.lang} {
+// CHECK-NEXT:    function.def @named_array_result
+// CHECK-SAME:    (%[[VAL_0:[0-9a-zA-Z_\.]+]]: !array.type<2 x index>, %[[VAL_1:[0-9a-zA-Z_\.]+]]: !array.type<2 x !felt.type>)
+// CHECK-SAME:    -> (!array.type<2 x index> {function.res_name = "out.lhs"}, !array.type<2 x !felt.type> {function.res_name = "out.rhs"}) {
```
0xddom: This may cause issues in `llzk-spec` since `out.lhs` is not a valid identifier. I'm not sure what the best solution would be. @iangneal thoughts?

test/Transforms/PodToScalar/ragged_nested_array_len.llzk
```diff
@@ -0,0 +1,63 @@
+// RUN: llzk-opt -split-input-file -llzk-pod-to-scalar -verify-diagnostics %s
+
+// COM: Regression test: reading one POD element from a flattened array-of-POD must reject nested
+// COM: dynamic or affine array leaves until the lowering can preserve per-element shape witnesses.
```
0xddom: I don't fully understand this issue

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -178,15 +328,328 @@ template <typename T> static bool containsSplittablePodType(ValueTypeRange<T> ty
...
+
+/// Return the attribute name that marks a nested dynamic array leaf lacking per-element shape.
+inline static StringRef getRaggedDynamicNestedLeafAttrName() {
+  return "llzk.ragged_nested_dynamic_leaf";
+}
+
+/// Return the attribute name that marks a nested affine array leaf lacking per-element shape.
+inline static StringRef getRaggedAffineNestedLeafAttrName() {
+  return "llzk.ragged_nested_affine_leaf";
+}
```
0xddom: These can be constants instead of functions.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -178,15 +328,328 @@ template <typename T> static bool containsSplittablePodType(ValueTypeRange<T> ty
...
+    );
+  }
+
+  llvm_unreachable("record path cannot continue through a non-POD leaf");
+}
+
+/// Return the flattened leaf type addressed by `recordChain` within `type`.
+inline static Type getFlattenedTypeAlongPath(Type type, ArrayRef<StringAttr> recordChain) {
+  return getFlattenedTypeAlongPath(type, recordChain, false);
+}
```
0xddom:
```suggestion
static Type
getFlattenedTypeAlongPath(Type type, ArrayRef<StringAttr> recordChain, bool syntheticShapeCarrier = false) {
  if (recordChain.empty()) {
    return type;
  }

  if (PodType podTy = dyn_cast<PodType>(type)) {
    Type nextType = podTy.getRecordMap().lookup(recordChain.front().getValue());
    assert(nextType && "record path must exist in the containing POD");
    return getFlattenedTypeAlongPath(nextType, recordChain.drop_front(), syntheticShapeCarrier);
  }

  if (ArrayType arrTy = splittablePodArray(type)) {
    if (syntheticShapeCarrier && recordChain.size() == 1) {
      assert(
          isPodArrayShapeCarrierMarker(recordChain.front()) &&
          "synthetic shape carrier must use the reserved shape marker"
      );
      assert(needsPodArrayShapeCarrier(arrTy) && "shape marker requires an explicit array carrier");
      return getPodArrayShapeCarrierType(arrTy);
    }

    auto elemPodTy = llvm::cast<PodType>(arrTy.getElementType());
    Type nextType = elemPodTy.getRecordMap().lookup(recordChain.front().getValue());
    assert(nextType && "record path must exist in the POD array element type");
    return flattenArrayElementType(
        arrTy, getFlattenedTypeAlongPath(nextType, recordChain.drop_front(), syntheticShapeCarrier)
    );
  }

  llvm_unreachable("record path cannot continue through a non-POD leaf");
}
```

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,203 +679,3289 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
+  return initialized ? CommonArrayInstantiationStatus::inferred
+                     : CommonArrayInstantiationStatus::unavailable;
+}
+
+/// Generate `arith.constant` indices for one static array element position.
+static SmallVector<Value> genArrayIndexConstants(OpBuilder &bldr, Location loc, ArrayAttr index) {
+  SmallVector<Value> indices;
+  for (Attribute attr : index) {
+    assert(llvm::isa<IntegerAttr>(attr) && "array index must be an integer attribute");
+    indices.push_back(bldr.create<arith::ConstantOp>(loc, llvm::cast<IntegerAttr>(attr)));
```
0xddom: nit: this assert is redundant

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,203 +679,3289 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
+
+/// Return the type produced by selecting `numIndices` leading dimensions from `arrTy`.
+static Type getArraySelectionType(ArrayType arrTy, size_t numIndices) {
+  assert(numIndices <= arrTy.getDimensionSizes().size() && "cannot select past the array rank");
+  if (numIndices == arrTy.getDimensionSizes().size()) {
+    return arrTy.getElementType();
+  }
+  return ArrayType::get(arrTy.getElementType(), arrTy.getDimensionSizes().drop_front(numIndices));
+}
+
```
0xddom: Isn't this a method in `ArrayType`? If not is a good candidate.

0xddom: Also, is it used anywhere other than line 1309? If not, then makes more sense to only return a sub-array type since the check done inside this function is already done before calling it below.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,203 +679,3289 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
+  SmallVector<Value> indices = genArrayIndexConstants(bldr, loc, index);
+  return genArrayRead(bldr, loc, arrayRef, indices);
+}
+
+/// Create an `array.write` or `array.insert` for one concrete element or subarray.
+static void
+genArrayWrite(OpBuilder &bldr, Location loc, Value arrayRef, ArrayRef<Value> indices, Value value) {
+  Type t = arrayRef.getType();
+  assert(llvm::isa<ArrayType>(t) && "array access must target an array type");
+  ArrayType arrTy = llvm::cast<ArrayType>(t);
```
0xddom: nit: redundant assert

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,203 +679,3289 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
+  }
+  return bldr.create<ExtractArrayOp>(
+      loc, llvm::cast<ArrayType>(getArraySelectionType(arrTy, indices.size())), arrayRef, indices
+  );
+}
+
+inline static Value genArrayRead(OpBuilder &bldr, Location loc, Value arrayRef, ArrayAttr index) {
+  SmallVector<Value> indices = genArrayIndexConstants(bldr, loc, index);
+  return genArrayRead(bldr, loc, arrayRef, indices);
+}
```
0xddom: This kind of read op creation logic (and the write counterparts below) happens quite often. I think we could move then somewhere in the vicinity of the array dialect to avoid having to rewrite it in the future.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,202 +344,1611 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
+  PodArrayTypeConverter typeConverter;
 
   RewritePatternSet patterns(ctx);
-
-  patterns.add<SplitPodInMemberDefOp>(ctx, symTables, memberRepMap);
+  patterns.add<
+      SplitPodArrayNonDetOp, SplitPodArrayCreateArrayOp, SplitPodArrayReadArrayOp,
+      SplitPodArrayWriteArrayOp, SplitPodArrayExtractArrayOp, SplitPodArrayInsertArrayOp,
+      SplitPodArrayInFuncDefOp, SplitPodArrayInReturnOp, SplitPodArrayInCallOp,
+      SplitPodArrayInEmitEqualityOp, SplitPodArrayInEmitContainmentOp, SplitPodArrayLengthOp>(
```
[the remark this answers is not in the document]
0xddom: I think instead of handling those we should lower them into loops with another pass that should be called before this one if necessary.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,203 +679,3289 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
+  llvm_unreachable("record path cannot continue through a non-POD leaf");
+}
+
+/// Read a flattened POD leaf by following each record name in `recordChain`.
+inline static Value
+genReadAlongPath(OpBuilder &bldr, Location loc, Value podRef, const RecordChain &recordChain) {
+  return genReadAlongPath(
+      bldr, loc, podRef, ArrayRef(recordChain.nameList), recordChain.syntheticShapeCarrier
+  );
+}
```
0xddom, line 1577: A bunch of these helper functions could be methods of their related types. For instance this one could be a method of `RecordChain`. Would make following the code a bit easier.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,203 +679,3289 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
+    }
+    return rebuiltArray;
+  }
+
+  auto it = leafValues.find(RecordChain(recordChain));
+  assert(it != leafValues.end() && "missing flattened POD leaf value");
+  return it->second;
+}
+
+using VirtualPodLeafMap = DenseMap<RecordChain, Value>;
```
0xddom: Why is this `using` here when the function just above it uses the aliased type?

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,203 +679,3289 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
+using VirtualPodValueMap = DenseMap<Value, VirtualPodLeafMap>;
+
+/// Strip compatibility casts around a virtual POD placeholder.
```
0xddom, line 1667: What is a virtual POD placeholder?

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -216,203 +679,3289 @@ splitPodType(TypeCollection types, SmallVector<size_t> *originalIdxToSize = null
...
+  if (PodType lhsTy = splittablePod(peelUnifiableCasts(op.getLhs()).getType())) {
+    return lhsTy;
+  }
+  return splittablePod(peelUnifiableCasts(op.getRhs()).getType());
+}
+
+/// Set the insertion point so a helper op dominates every use of `value`.
+static void setInsertionPointAfterValueDefinition(Value value, OpBuilder &bldr) {
+  if (Operation *defOp = value.getDefiningOp()) {
+    bldr.setInsertionPointAfter(defOp);
```
0xddom, line 1757: If the goal is to dominate `value` shouldn't the insertion point be before the defining op?

0xddom (dismissed):
The `PodToScalarPass.cpp` file is helper functions galore. The code is hard to follow and I think it could be organized a bit better.

test/Transforms/PodToScalar/constraints_and_whole_pod_uses.llzk
tim-hoffman: It's handled via the `SymbolTable::insert()` but I'll add test coverage.

test/Transforms/PodToScalar/constraints_and_whole_pod_uses.llzk
tim-hoffman: Does `@nested_if_external_new`, `@if_result_and_external_write`, or `@if_write_then_whole_pod_consume` cover the scenario you're thinking of?

test/Transforms/PodToScalar/constraints_and_whole_pod_uses.llzk
tim-hoffman: Oh, it seems I missed several `unrealized_conversion_cast` that were left behind and should be handled (if possible).

test/Transforms/PodToScalar/constraints_and_whole_pod_uses.llzk
tim-hoffman: Yep. I thought I fixed that.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
tim-hoffman: There are a handful that made sense to move but this one and `getFlattenedTypeAlongPath()` would require pre-defining the callees and/or moving a bunch of stuff around so the declaration ordering is correct and I don't think that's worth the bloat/effort.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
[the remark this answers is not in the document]
tim-hoffman: That does sound better.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
tim-hoffman: Good catch. I'll add it to `ArrayType`. I see a similar pattern in `lib/Dialect/Array/IR/Ops.cpp`, `lib/Dialect/Bool/IR/Ops.cpp`, and `lib/Analysis/SourceRef.cpp`.

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
tim-hoffman: "dominates every use of `value`" so inserting right after the definition of value ensures that it dominates all uses

include/llzk/Dialect/Array/IR/Types.td
```diff
@@ -118,6 +118,9 @@ def LLZK_ArrayType
     /// Required by the ShapedTypeInterface interface.
     inline bool hasRank() const { return true; }
 
+    /// Return the type produced by selecting/removing `numIndices` leading dimensions.
```
0xddom: I'd add that passing exactly the number of dimensions returns the element type.

test/Transforms/PodToScalar/constraints_and_whole_pod_uses.llzk
0xddom: Makes sense that is handled via the symbol table. My main worry was making sure the right name was used for creating the read/write ops.

test/Transforms/PodToScalar/constraints_and_whole_pod_uses.llzk
0xddom: It's _fine_ to leave some unrealized conversions if passes down the line can clear them up

test/Transforms/PodToScalar/constraints_and_whole_pod_uses.llzk
0xddom: Yeah, I saw those after I wrote that comment. That's what I had in mind

test/Transforms/PodToScalar/function_calls_with_pod.llzk
tim-hoffman: I think Ian and I discussed this before and decided this would be fine.

tim-hoffman:
@codex review

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -288,131 +1953,1836 @@ inline static void baseTargetSetup(ConversionTarget &target) {
...
+          rhsTy ? adaptor.getRhs() : ValueRange(rhsCompatibleConvertedValues), op.getLoc(), rewriter
+      );
+
+      for (size_t dim = 0, rank = shapeCheckTy.getDimensionSizes().size(); dim < rank; ++dim) {
+        Value dimVal = rewriter.create<arith::ConstantOp>(
+            op.getLoc(), rewriter.getIndexAttr(llzk::checkedCast<int64_t>(dim))
+        );
+        Value lhsLen = rewriter.create<ArrayLengthOp>(op.getLoc(), lhsShapeSource, dimVal);
+        Value rhsLen = rewriter.create<ArrayLengthOp>(op.getLoc(), rhsShapeSource, dimVal);
+        rewriter.create<constrain::EmitEqualityOp>(op.getLoc(), lhsLen, rhsLen);
```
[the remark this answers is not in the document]
tim-hoffman: LLZKPolyLoweringPass should be updated to handle this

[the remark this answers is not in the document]
tim-hoffman: https://github.com/project-llzk/llzk-lib/issues/642

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -288,131 +1953,1836 @@ inline static void baseTargetSetup(ConversionTarget &target) {
...
+      Value dimVal = rewriter.create<arith::ConstantOp>(
+          loc, rewriter.getIndexAttr(llzk::checkedCast<int64_t>(dim))
+      );
+      Value dimLen = rewriter.create<ArrayLengthOp>(loc, shapeCarrier, dimVal);
+
+      Value nonNegative = rewriter.create<arith::CmpIOp>(loc, arith::CmpIPredicate::sge, idx, zero);
+      rewriter.create<constrain::EmitEqualityOp>(loc, nonNegative, trueVal);
+
+      Value inRange = rewriter.create<arith::CmpIOp>(loc, arith::CmpIPredicate::slt, idx, dimLen);
+      rewriter.create<constrain::EmitEqualityOp>(loc, inRange, trueVal);
```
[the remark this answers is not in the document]
tim-hoffman: https://github.com/project-llzk/llzk-lib/issues/642

tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

lib/Dialect/POD/Transforms/PodToScalarPass.cpp
```diff
@@ -178,640 +361,4938 @@ template <typename T> static bool containsSplittablePodType(ValueTypeRange<T> ty
...
+    }
+    if (promotedSignatures.empty()) {
+      return success();
+    }
+    if (failed(updateCallsForPromotedFunctionArgCasts(modOp, symTables, promotedSignatures))) {
+      return failure();
+    }
+    if (promotionRounds % 64 == 0) {
+      llvm::outs() << "function argument cast promotion has run " << promotionRounds
+                   << " rounds without reaching a fixpoint; continuing...\n";
```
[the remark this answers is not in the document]
tim-hoffman: won't fix

tim-hoffman:
@codex review

tim-hoffman:
@codex review

shankarapailoor (approved):
LGTM

PR #566, opened 2026-06-23
Add support for `@product` functions in the CAPI

iangneal:


lib/Dialect/Function/IR/Ops.cpp
```diff
@@ -1167,7 +1173,7 @@ FailureOr<SymbolLookupResult<FuncDefOp>> CallOp::getCalleeTarget(SymbolTableColl
 
 StructType CallOp::getSingleResultTypeOfCompute() {
   assert(calleeIsCompute() && "violated implementation pre-condition");
-  return getIfSingleton<StructType>(getResultTypes());
+  return getSingleResultTypeOfWitnessGen();
```
tim-hoffman: the one downside to this is executing two assert checks vs one when `calleeIsCompute()` implies `calleeContainsWitnessGen()`

tim-hoffman (approved):
LGTM, just one nit

tim-hoffman (approved):
LGTM

PR #572, opened 2026-06-26
Respect table offsets in duplicate read/write elimination

### LINKED ISSUE #569: `--llzk-duplicate-read-write-elim` ignores `tableOffset` when reusing `struct.readm`
### Summary

`--llzk-duplicate-read-write-elim` can reuse a `struct.readm` for the same member across different `tableOffset` values.

### Reproduction

Reproducer run:
https://github.com/Kuhai9801/llzk-lib/actions/runs/28180228869

```mlir
struct.member @cell : !felt.type {column}

function.def @constrain(%self: !struct.type<@Rows>) {
  %prev = struct.readm %self[@cell] : !struct.type<@Rows>, !felt.type {tableOffset = -1 : index}
  %curr = struct.readm %self[@cell] : !struct.type<@Rows>, !felt.type {tableOffset = 0 : index}
  constrain.eq %prev, %curr : !felt.type, !felt.type
  function.return
}
```

Run: `llzk-opt --llzk-duplicate-read-write-elim repro.llzk`

Actual output excerpt:

```mlir
%0 = struct.readm %arg0[@cell] : <@Rows>, !felt.type {tableOffset = -1 : index}
constrain.eq %0, %0 : !felt.type, !felt.type
```

### Expected

The two reads remain distinct because they use different row offsets.

### Actual

The current-row read is replaced by the prior-row read, and the resulting constraint compares the same SSA value to itself.

### Notes

The relevant pass is in `lib/Transforms/LLZKRedundantReadAndWriteEliminationPass.cpp`.

1sgtpepper:
Fixes #569.

## Summary
`llzk-duplicate-read-write-elim` now includes struct member table offsets and affine operands in the access identity, so reads from different rows or affine operands are not conflated.

## Changes
- Include `tableOffset`, affine operand group metadata, and translated affine operands in member-access keys.
- Treat omitted `tableOffset` as the current row.
- Keep constant nonzero row reads distinct across current-row writes.
- Invalidate symbolic and affine offset reads when a current-row write may alias them.
- Add regression coverage for different, same, default, write, and affine offset cases.

## Testing
- `git diff --check`
- `git-clang-format --diff upstream/main -- include/llzk/Transforms/LLZKTransformationPasses.td lib/Transforms/LLZKRedundantReadAndWriteEliminationPass.cpp`
- GitHub Actions: Clang/GCC lit and unit tests, Nix debug/release checks, docs, changelog, format, publish-tests
- Regression covered by `test/Transforms/RedundantAndUnusedElim/redundant_read_write_table_offset.llzk`

### DIFF
--- a/test/Transforms/RedundantAndUnusedElim/redundant_read_write_table_offset.llzk
+++ b/test/Transforms/RedundantAndUnusedElim/redundant_read_write_table_offset.llzk
@@ -1,0 +1,356 @@
+// RUN: llzk-opt -split-input-file -llzk-duplicate-read-write-elim %s 2>&1 | FileCheck %s
+
+module attributes {llzk.lang} {
+  struct.def @DifferentOffsets {
+    struct.member @cell : !felt.type {column}
+
+    function.def @compute() -> !struct.type<@DifferentOffsets> {
+      %self = struct.new : !struct.type<@DifferentOffsets>
+      function.return %self : !struct.type<@DifferentOffsets>
+    }
+
+    function.def @constrain(%self: !struct.type<@DifferentOffsets>) {
+      %prev = struct.readm %self[@cell] : !struct.type<@DifferentOffsets>, !felt.type {tableOffset = -1 : index}
+      %curr = struct.readm %self[@cell] : !struct.type<@DifferentOffsets>, !felt.type {tableOffset = 0 : index}
+      constrain.eq %prev, %curr : !felt.type
+      function.return
+    }
+  }
+}
+
+// CHECK-LABEL: struct.def @DifferentOffsets {
+// CHECK: function.def @constrain(%[[SELF:.*]]: !struct.type<@DifferentOffsets>)
+// CHECK-NEXT: %[[PREV:.*]] = struct.readm %[[SELF]][@cell] : <@DifferentOffsets>, !felt.type {tableOffset = -1 : index}
+// CHECK-NEXT: %[[CURR:.*]] = struct.readm %[[SELF]][@cell] : <@DifferentOffsets>, !felt.type {tableOffset = 0 : index}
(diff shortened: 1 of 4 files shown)


iangneal:
@codex review

test/Transforms/RedundantAndUnusedElim/redundant_read_write_table_offset.llzk
```diff
@@ -0,0 +1,356 @@
...
+}
+
+// CHECK-LABEL: struct.def @WriteOffsets {
+// CHECK: function.def @compute(%[[VALUE:.*]]: !felt.type) -> !struct.type<@WriteOffsets>
+// CHECK-NEXT: %[[SELF:.*]] = struct.new : <@WriteOffsets>
+// CHECK-NEXT: %[[PREV:.*]] = struct.readm %[[SELF]][@cell] : <@WriteOffsets>, !felt.type {tableOffset = -1 : index}
+// CHECK-NEXT: struct.writem %[[SELF]][@cell] = %[[VALUE]] : <@WriteOffsets>, !felt.type
+// CHECK-NEXT: struct.writem %[[SELF]][@first] = %[[PREV]] : <@WriteOffsets>, !felt.type
+// CHECK-NEXT: struct.writem %[[SELF]][@second] = %[[PREV]] : <@WriteOffsets>, !felt.type
+// CHECK-NEXT: struct.writem %[[SELF]][@current] = %[[VALUE]] : <@WriteOffsets>, !felt.type
```
iangneal, line 117: Can we change these to full function checks instead of partial function checks? Just because this method of checking doesn't exclude additional reads/writes after the above CHECKS

iangneal (changes requested):
Just one small request about test formatting, so we can match the style of lit tests throughout the rest of the repo.

iangneal (approved):
LGTM, thank you!

PR #573, opened 2026-06-29
Add ram dialect to polymorphic transforms

matthew-a-klein:
The polymorphic transform passes treat any op from a dialect not in `newBaseTarget`'s legal-dialect list as illegal. `ram` ops were missing, so any module containing `ram.load`/`ram.store` would fail full-conversion through these passes even though there is nothing for the polymorphic conversion to do to them


changelogs/unreleased/add-ram-dialect-flattening.yaml
```diff
@@ -0,0 +1,2 @@
+fixed:
+  - Allow `ram` dialect ops to pass through polymorphic transforms
```
iangneal: Nit: to silence GitHub warnings on missing newlines
```suggestion
  - Allow `ram` dialect ops to pass through polymorphic transforms
  
```

iangneal (changes requested):
Overall LGTM, just one nit

changelogs/unreleased/add-ram-dialect-flattening.yaml
matthew-a-klein: Ok, looks like you have addressed this and will merge once CI finishes?

iangneal:
@matthew-a-klein thank you!

PR #575, opened 2026-07-01
struct.readm/writem loop alignment

raghav198:
A common pattern for structs with loops is 

```
func @compute(...) {
    %arr = array.new ...
    scf.for {
        ...
    }
    struct.writem %self[@arr] = %arr
}
```

```
func @constrain(...) {
    %arr = struct.readm %self[@arr]
    scf.for {
        ...
    }
}
```

Naively aligning the two loops just by moving one up/down to the other will violate one of these dependences (e.g., write an empty array before its been populated inside the fused loop).

This PR fixes loop alignment so that when the compute/constrain loops are aligned, the `struct.writem` is kept after the aligned loop and the `struct.readm` is kept before the aligned loop, and fails to align if this is not possible.


iangneal (changes requested):
Needs a changelog and regression tests.

lib/Transforms/LLZKFuseProductLoopsPass.cpp
```diff
@@ -118,6 +120,100 @@ static inline bool canLoopsBeFused(scf::ForOp a, scf::ForOp b) {
...
 
+static StringRef getProductSource(Operation *op) {
+  if (auto attr = op->getAttrOfType<StringAttr>(PRODUCT_SOURCE)) {
+    return attr.getValue();
+  }
+  return {};
+}
+
+static Operation *getTopLevelAncestorInBlock(Operation *op, Block *block) {
+  while (op && op->getBlock() != block) {
```
tim-hoffman: Why is the block check inequality rather than equality?

lib/Transforms/LLZKFuseProductLoopsPass.cpp
```diff
@@ -118,6 +120,100 @@ static inline bool canLoopsBeFused(scf::ForOp a, scf::ForOp b) {
...
+  }
+
+  if (between.empty() && firstLoop->getNextNode() != secondLoop) {
+    return std::nullopt;
+  }
+
+  return between;
+}
+
+static bool canPrepareForFusion(
```
tim-hoffman: Could you add doc comments on these functions?

lib/Transforms/LLZKFuseProductLoopsPass.cpp
```diff
@@ -43,7 +45,7 @@ constexpr int INDEX_WIDTH = 64;
 static inline bool isConstOrStructParam(Value val) {
   // TODO: doing arithmetic over constants should also be fine?
   return val.getDefiningOp<arith::ConstantIndexOp>() ||
-         val.getDefiningOp<polymorphic::ConstReadOp>();
+         val.getDefiningOp<polymorphic::ConstReadOp>() || val.getDefiningOp<felt::FeltConstantOp>();
```
iangneal: Nit:
```suggestion
         return llvm::isa<arith::ConstantIndexOp, polymorphic::ConstReadOp, felt::FeltConstantOp>(val.getDefiningOp());
```

test/Transforms/FuseProductLoops/no_fuse_sink.llzk
```diff
@@ -0,0 +1,48 @@
...
+                %my_arr_i = array.read %my_arr[%i] : !A, !F {product_source = "constrain"}
+                constrain.eq %arr1_i, %my_arr_i : !F {product_source = "constrain"}
+            } {product_source = "constrain"}
+
+            function.return %self : !S
+        }
+    }
+}
+
+// CHECK-LABEL: scf.for
```
iangneal: I still prefer having the full output in the checks, scripts/generate-test-checks.py is good for this.

iangneal (commented):
A couple of nits

iangneal (approved):

Issue #544, opened 2026-06-14
(proposal): preserve freshness-scoped polynomial auxiliaries before storage materialization


1sgtpepper:
### Summary

This proposal is the longer-term architecture behind #542.

The core design rule is:

> Polynomial lowering should introduce compiler-generated existential SSA auxiliaries with an explicit dynamic freshness domain. It should not immediately materialize those auxiliaries as `struct.member` storage.

A later materialization layer should decide how each auxiliary is represented for a particular target: scalar component member, backend-private signal, static unrolled member, array element, column/trace cell, product-program value, deterministic witgen value, or diagnostic rejection.

The intent is not to make #542 wait for a large rewrite. The immediate #542 fix can be a conservative rejection in the current scalar-member path. This proposal explains the architecture that keeps that fix from becoming a chain of special cases.

### Motivation

Polynomial lowering is an existential-introduction transform.

At the algebraic level, it changes:

```text
P[E]
```

into:

```text
exists a. (a = E) and P[a]
```

That is only sound when the scope of `exists a` matches the dynamic execution of `E`.

For a component-invariant expression, this may be sound:

```text
exists a_component.
  a_component = E
  and P[a_component]
```

For an expression that depends on a loop-carried value, the required meaning is different:

```text
forall i.
  exists a_i.
    a_i = E(i)
    and P_i[a_i]
```

The unsound form is:

```text
exists a_component.
  forall i.
    a_component = E(i)
    and P_i[a_component]
```

That form adds constraints not present in the source program unless `E(i)` is proven identical for every iteration.

#542 demonstrates this exact failure. A loop-local expression depending on an `scf.for` loop-carried region argument is lowered to one component-scope auxiliary member. The source constraint is tautological, but the lowered constraints can require the same generated member to equal different per-iteration values.

### Concrete current behavior

The current pass contract already points to the semantic mismatch. The pass documentation says polynomial lowering rewrites constraints into an observationally equivalent bounded-degree system, but also says high-degree subexpressions are factored into auxiliary struct members:
https://github.com/project-llzk/llzk-lib/blob/aced31cd0609fc26eaf0dbb0b608f7043c8d93b5/include/llzk/Transforms/LLZKTransformationPasses.td#L66-L79

The implementation creates auxiliary `struct.member`s during expression factoring:
[...]

PR #579, opened 2026-07-02
Add placement-sensitive llzk.aux operation

1sgtpepper:
## Summary

Adds `llzk.aux` as a felt-only, placement-sensitive IR anchor for compiler-generated polynomial auxiliaries.

The op is intentionally conservative: it has no purity, speculation, or constant-like traits, and it does not add a canonicalizer. This keeps distinct generated auxiliaries from being treated as interchangeable values before later lowering/materialization work proves that would be legal.

Related to #544.


1sgtpepper:
Closing this draft with #544. Further investigation of the current backend and lowering pipelines found no path that would produce or consume `llzk.aux`: R1CS flattens before polynomial lowering, while PCL lowers arithmetic directly. Keeping the operation would add unused IR surface, so I’m closing this and would revisit it only around a concrete end-to-end consumer.

PR #593, opened 2026-07-08
fix: flattening pass crash from felt const in array dim

tim-hoffman:


test/Transforms/Flattening/felt_type_poly_expr_as_dim.llzk
```diff
@@ -0,0 +1,35 @@
...
+  poly.template @T {
+    poly.param @n
+    poly.expr @n_Add_1 {
+      %felt_const_1 = felt.const  1 : <"bn128">
+      %0 = poly.read_const @n : !felt.type<"bn128">
+      %1 = felt.add %0, %felt_const_1 : !felt.type<"bn128">, !felt.type<"bn128">
+      poly.yield %1 : !felt.type<"bn128">
+    }
+    struct.def @T {
+      struct.member @r : !array.type<@n_Add_1 x !felt.type<"bn128">>
```
iangneal, line 13: Should this be legal? I would think that, rather than a crash, this should report a type error. Since the array dimensions need to be `index` type, why not only have this legal if the `poly.expr` yields an `index` type? This kinda makes `poly.expr` a special case where we implicitly convert `felt.type` to `index`, which is not present in other cases.

tim-hoffman: That's fair. I need to change what's generated on the circom side instead and look into producing a verification error on the LLZK side.

tim-hoffman:
Closing in favor of combination of https://github.com/project-llzk/llzk-lib/pull/595 and https://github.com/project-llzk/circom/pull/444

PR #598, opened 2026-07-10
Update `SourceRef` printing

iangneal:
- Use `function.arg_name` and `function.res_name` for string-ification, if available
- Use `.` syntax for accessing struct members and pod records to reduce verbosity and differentiate between array indices


iangneal:
@codex review

lib/Dialect/Function/IR/Ops.cpp
```diff
@@ -119,6 +119,23 @@ static LogicalResult verifyArgOrResNameAttrs(
...
+  if (auto dictAttr = llvm::dyn_cast<DictionaryAttr>(attrs[index])) {
+    Attribute attr = dictAttr.get(attrName);
+    if (!attr) {
+      return std::nullopt;
+    }
+    if (auto nameAttr = llvm::dyn_cast<StringAttr>(attr)) {
+      return nameAttr;
+    }
+  }
+  return std::nullopt;
```
tim-hoffman, line 133: You can avoid the first null check by using `dyn_cast_if_present`

tim-hoffman (approved):
LGTM, just one nit

lib/Dialect/Function/IR/Ops.cpp
iangneal: Aw man, I thought I fixed all these

tim-hoffman (approved):

PR #599, opened 2026-07-10
Add `function.res_name` functions to the CAPI

iangneal:


lib/CAPI/Dialect/Function.cpp
```diff
@@ -80,6 +80,23 @@ void llzkFunction_FuncDefOpSetArgName(MlirOperation op, unsigned index, MlirStri
   llvm::cast<FuncDefOp>(unwrap(op)).setArgName(index, unwrap(name));
 }
 
+bool llzkFunction_FuncDefOpHasResNameAttr(MlirOperation op, unsigned index) {
+  return llvm::cast<FuncDefOp>(unwrap(op)).hasResName(index);
+}
```
tim-hoffman: This one is redundant per one that's auto-generated:
```
bool llzkFunction_FuncDefOpHasResName(MlirOperation inp, unsigned index) {
  return llvm::cast<FuncDefOp>(unwrap(inp)).hasResName(index);
}
```

tim-hoffman: The same is true for `llzkFunction_FuncDefOpHasArgNameAttr` (which already exists) per:
```
bool llzkFunction_FuncDefOpHasArgName(MlirOperation inp, unsigned index) {
  return llvm::cast<FuncDefOp>(unwrap(inp)).hasArgName(index);
}
```

tim-hoffman (approved):
LGTM

PR #605, opened 2026-07-14
Align APInt storage hashing with numeric equality

### LINKED ISSUE #615: Different-width APInt attributes can alias after a hash collision
`APIntParameter` hashes include bit width but compares using `APInt::isSameValue`, which ignores it. If MLIR's truncated storage hashes collide, same-valued APInts with different widths alias and the later attribute inherits the first width.

CI repro: value 909, widths 2499/2718; the 2718-bit attribute reports 2499 bits ([failing GCC job](https://github.com/1sgtpepper/llzk-lib/actions/runs/29463037085/job/87510348995)). The width-aware comparator passes the same test ([fixed GCC job](https://github.com/1sgtpepper/llzk-lib/actions/runs/29463127688/job/87510609265)).

Related: #605 and #32.

1sgtpepper:
Fixes #615.

Make generic `APIntParameter` storage hashing match its width-independent `isSameValue` equality. Keep SMT bit-vectors width-sensitive because their APInt width determines the type.

Validation:
- cross-width numeric APInts reuse storage with equal hashes
- equal-valued SMT bit-vectors with different widths remain distinct
- same-width SMT bit-vectors reuse storage
- CI

### DIFF
--- a/include/llzk/Dialect/LLZK/IR/AttributeHelper.td
+++ b/include/llzk/Dialect/LLZK/IR/AttributeHelper.td
@@ -12,10 +12,19 @@
 
 include "mlir/IR/AttrTypeBase.td"
 
-// APInts that allows comparison involving different bitwidths.
+// APInts whose numeric value is independent of bitwidth.
 class APIntParameter<string desc = "">
+    : AttrOrTypeParameter<"::llzk::APIntValue", desc, "const ::llvm::APInt &"> {
+  let parser = "::llzk::parseAPIntValue($_parser)";
+  let printer = "$_printer << $_self";
+}
+
+// APInts whose bitwidth is part of their identity. Check it before operator==,
+// which requires equal widths.
+class WidthSensitiveAPIntParameter<string desc = "">
     : AttrOrTypeParameter<"::llvm::APInt", desc> {
-  let comparator = "::llvm::APInt::isSameValue($_lhs, $_rhs)";
+  let comparator =
+      "$_lhs.getBitWidth() == $_rhs.getBitWidth() && $_lhs == $_rhs";
 }
 
 #endif // LLZK_ATTRIBUTE_HELPER
--- a/include/llzk/Dialect/SMT/IR/SMTAttributes.td
+++ b/include/llzk/Dialect/SMT/IR/SMTAttributes.td
@@ -46,7 +46,7 @@ def BitVectorAttr
     present).
   }];
 
-  let parameters = (ins APIntParameter<"">:$value);
+  let parameters = (ins WidthSensitiveAPIntParameter<"">:$value);
 
   let hasCustomAssemblyFormat = true;
   let genVerifyDecl = true;
--- a/unittests/IR/SMTAttributeTests.cpp
+++ b/unittests/IR/SMTAttributeTests.cpp
@@ -11,84 +11,10 @@
 
 #include "llzk/Dialect/Felt/IR/Attrs.h"
 #include "llzk/Dialect/LLZK/IR/AttributeHelper.h"
-#include "llzk/Dialect/SMT/IR/SMTAttributes.h"
 
 #include <llvm/ADT/APInt.h>
-#include <llvm/ADT/DenseMap.h>
-#include <llvm/ADT/Hashing.h>
-
-#include <cstdint>
-#include <optional>
-
-using namespace llzk::smt;
 
 class SMTAttributeTests : public LLZKTest {};
-
-namespace {
-
-struct APIntStorageCollision {
-  uint64_t value;
-  unsigned narrowWidth;
-  unsigned wideWidth;
-  unsigned hash;
-};
-
-std::optional<APIntStorageCollision> findAPIntStorageCollision() {
-  constexpr unsigned minWidth = 64;
-  constexpr unsigned maxWidth = 4096;
-  constexpr uint64_t maxValue = 100000;
-
-  llvm::DenseMap<uint64_t, unsigned> widthByHash;
-  widthByHash.reserve(maxWidth - minWidth + 1);
-
-  for (uint64_t value = 0; value < maxValue; ++value) {
-    widthByHash.clear();
-    for (unsigned width = minWidth; width <= maxWidth; ++width) {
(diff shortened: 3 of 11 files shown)


include/llzk/Dialect/LLZK/IR/AttributeHelper.td
```diff
@@ -12,10 +12,10 @@
 
 include "mlir/IR/AttrTypeBase.td"
 
-// APInts that allows comparison involving different bitwidths.
+// APInts keyed by bit width and value.
```
iangneal: The whole point of this parameter was to allow for comparison of APInts regardless of bit width. You've now changed the semantics of this type of parameter. Why have you changed this? What does this fix?

iangneal (changes requested):
I seriously question the purpose of this PR. Can you explain what this is supposed to fix?

changelogs/unreleased/apint-storage-key-equality.yaml
iangneal, line 1: This changelog fails the check since it isn't named after your branch `pcl-lowering-portable-checks`

iangneal (changes requested):

include/llzk/Dialect/LLZK/IR/AttributeHelper.td
1sgtpepper: Thanks, you're right that this changes the parameter's semantics, so I'm moving the PR back to draft while I confirm the intended contract.

iangneal:
@1sgtpepper my question still stands: what are you trying to fix in this PR? I am unaware of any bug related to this parameter.

include/llzk/Dialect/LLZK/IR/AttributeHelper.td
iangneal: > Thanks, you're right that this changes the parameter's semantics, so I'm moving the PR back to draft while I confirm the intended contract.

I'd like to give you some feedback on this response and explain why this (AI) response isn't particularly helpful when replying to feedback we give on pull requests:

1. I asked you two questions and you answered neither of them.
2. The following:
> so I'm moving the PR back to draft while I confirm the intended contract.

is a bit nonsensical. The "intended contract" was in the comment you edited: `// APInts that allows comparison involving different bitwidths.` Also, there's no bugs or issues filed with regard to this, and no changes have been requested, so as far as I'm aware, there is no other "intended contract" anywhere.

Going forward, let's aim to address reviewer comments and questions directly and clearly in replies, and ultimately this will help both of us, as it will help the reviewers review faster and it will reduce the back-and-forth of questions you need to reply to to get your changes integrated.

1sgtpepper: You’re right—sorry. I had a local reproducer but failed to include or explain it. See #615.

1sgtpepper: I changed it because `APIntParameter`’s hash includes bit width while `isSameValue` does not. On a truncated-hash collision, different-width attributes can alias and inherit the wrong width; #615 reproduces it in CI.

1sgtpepper: Apologies again—the reply was posted prematurely while I was drafting a response to your questions.

iangneal: > I changed it because APIntParameter’s hash includes bit width while isSameValue does not. On a truncated-hash collision, different-width attributes can alias and inherit the wrong width;

Ah okay, thank you for the explanation. I think the correct solution, however, should be to remove bit width from `APIntParameter`'s hash value rather than adding bit width to the comparison check.

1sgtpepper: Agreed. Since SMT bit-vector width is part of the type, should this PR keep the width-sensitive SMT parameter and move the generic `APIntParameter` hash fix to a follow-up?

include/llzk/Dialect/SMT/IR/SMTAttributes.td
```diff
@@ -46,7 +46,7 @@ def BitVectorAttr
     present).
   }];
 
-  let parameters = (ins APIntParameter<"">:$value);
+  let parameters = (ins WidthSensitiveAPIntParameter<"">:$value);
```
iangneal, line 49: See this doesn't actually need to be width sensitive, since this parameter encodes the size of the bitvector itself. So by saying this is width-sensitive, you're saying that the width of the width is important, which it still isn't.

iangneal (changes requested):

include/llzk/Dialect/SMT/IR/SMTAttributes.td
1sgtpepper: Thanks, Ian. I may be missing a distinction, but this APInt seems to hold the bit-vector value itself, and `getValue().getBitWidth()` becomes the `BitVectorType` width. So I think 4-bit and 8-bit `5` need distinct identities. I kept generic APInts width-independent and limited width sensitivity to `BitVectorAttr`, but I am happy to adjust if I have misunderstood.

iangneal: Oh no, I believe you're right, I just misread the example.

unittests/IR/SMTAttributeTests.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+  return std::nullopt;
+}
+
+} // namespace
+
+TEST_F(SMTAttributeTests, BitVectorStorageDoesNotAliasCollidingWidths) {
+  // Debug LLVM builds seed hashes per process, so fixed collision values are
+  // not portable. Find a collision in-process to exercise storage equality.
+  std::optional<APIntStorageCollision> collision = findAPIntStorageCollision();
```
iangneal: This test seems potentially flaky and expensive. I think we can remove this one and just keep the one below.

iangneal (changes requested):
Overall LGTM, just one comment about the unit tests

iangneal (approved):
LGTM!

PR #614, opened 2026-07-15
fix: read_const cannot target type variables

tim-hoffman:


include/llzk/Dialect/Polymorphic/IR/Ops.td
iangneal: Summary should also be changed to "read value of a template parameter"

iangneal (changes requested):
My broader question: why this restriction?

tim-hoffman:
> My broader question: why this restriction?

`poly.param` with `tvar` type means that the symbol should only be used in places where a type is allowed:

https://github.com/project-llzk/llzk-lib/blob/f6abca33ef6697b9c6acb53581cfc0409073e398/include/llzk/Dialect/Polymorphic/IR/Ops.td#L158-L159

Types are not first-class values in LLZK so you can't grab the type from a parameter and store it in an SSA value.

iangneal:
> > My broader question: why this restriction?
> 
> `poly.param` with `tvar` type means that the symbol should only be used in places where a type is allowed:
> 
> https://github.com/project-llzk/llzk-lib/blob/f6abca33ef6697b9c6acb53581cfc0409073e398/include/llzk/Dialect/Polymorphic/IR/Ops.td#L158-L159
> 
> Types are not first-class values in LLZK so you can't grab the type from a parameter and store it in an SSA value.

This seems to contradict the documentation for `poly.tvar` itself:

https://github.com/project-llzk/llzk-lib/blob/f6abca33ef6697b9c6acb53581cfc0409073e398/include/llzk/Dialect/Polymorphic/IR/Types.td#L30-L43

Based on this example, an instantiation of `@TA::@A<[!felt.type]>` turns

```
poly.template @TA {
    poly.param @Ty
    struct.def @A {
      member @x : !array.type<2 x !poly.tvar<@Ty>>
      ...
    }
  }
```

into 

```
poly.template @TA {
    struct.def @A {
      member @x : !array.type<2 x !felt.type>
      ...
    }
  }
```

Correct? So why does 

```
%foo = poly.read_const : !poly.tvar<@Ty>
```

Not become

```
%foo = poly.read_const : !felt.type
```

Which would be legal?

tim-hoffman:
> Correct? So why does
> 
> 
> ```
> %foo = poly.read_const : !poly.tvar<@Ty>
> ```
> 
> 
>     
>   
> 
> Not become
> 
> 
> ```
> %foo = poly.read_const : !felt.type
> ```
> 
> 
>     
>   
> 
> Which would be legal?

You left out the symbol name. Valid syntax would be:
`%foo = poly.read_const @Ty : !poly.tvar<@Ty>`

The first position `@Ty` expects to be replaced with a value (not a type since a Type cannot be a value). The second position `@Ty` expects to be replaced with a Type. Hence, a contradiction because a single parameter cannot be both a value and a type.

Consider the explicit template parameter list syntax:
`function.call @Template::@f<[X]>()`
What should provided for X? The value `27` or the type `index`?

You can achieve what you're wanting but you need two template params:
`%foo = poly.read_const @N : !poly.tvar<@Ty>`
I added a test case showing how to do this.

iangneal:
Oh oh I see what you mean, I misinterpreted "targeting type variables" meaning the couldn't be used at all in the `poly.read_const` expression. Mostly my own mis-reading, but adding an example of what you said above would be helpful.

tim-hoffman:
> Oh oh I see what you mean, I misinterpreted "targeting type variables" meaning the couldn't be used at all in the `poly.read_const` expression. Mostly my own mis-reading, but adding an example of what you said above would be helpful.

I added this test as example: https://github.com/project-llzk/llzk-lib/pull/614/changes#diff-b3ac6f92cf25ebbb84d268234413bc1d2ba9df76b219e82b66570fc01a4e9ddb
Let me know if add clarifications/examples in doc/comments would be helpful anywhere a well.

include/llzk/Dialect/Polymorphic/IR/Ops.td
```diff
@@ -225,13 +225,12 @@ def LLZK_YieldOp
...
-  let summary = "read value of a struct parameter";
+  let summary = "read value of a template parameter";
   let description = [{
-    This operation reads the value from the named constant parameter of
-    the struct/component in which this op appears. The op itself puts
-    some restriction on the type of this value, but leaves it to a later
-    type-checking pass to ensure the struct parameters are instantiated
-    with types matching the uses of the parameter within the struct.
+    This operation reads the value from the named constant parameter of the template in
+    which this op appears. The op itself puts some restriction on the type of this value,
```
iangneal: I do find "some restriction" to be vague, can we be a little more precise here?

iangneal (dismissed):
Maybe just clarify the const_read docs a bit but otherwise LGTM

iangneal (approved):
LGTM!

PR #621, opened 2026-07-17
Fix instability in PCL test output by dropping the int constant bitwidth from CHECKs

tim-hoffman:


iangneal (approved):
I filled an issue for this for the broader issue that this represents, being that the bit widths shouldn't cause test flakiness, but I think this fix is fine in the interim.

PR #624, opened 2026-07-17
Add support for PODs in the interval analysis

iangneal:


unittests/CAPI/Dialect/Array.cpp
```diff
@@ -20,9 +20,13 @@
 #include <mlir-c/BuiltinTypes.h>
 #include <mlir-c/IR.h>
 
+#include <mlir/IR/BuiltinOps.h>
+
 #include <llvm/ADT/ArrayRef.h>
 #include <llvm/ADT/SmallVector.h>
 
+#include <array>
+
```
tim-hoffman: are these changes necessary?

tim-hoffman (approved):
LGTM, just if those new includes are not necessary they should be removed

unittests/CAPI/Dialect/Array.cpp
iangneal: Ooops, I accidentally included these from other debugging task that I meant to stash.

tim-hoffman (approved):

PR #627, opened 2026-07-21
PCL Improvements

0xddom:


backends/pcl/include/pcl/Dialect/IR/Ops.td
```diff
@@ -167,12 +202,14 @@ def PCL_AssertOp : PCL_Op<"assert", []> {
   let summary = "Assert that a boolean holds";
   let arguments = (ins PCL_BoolType:$cond);
   let assemblyFormat = "$cond attr-dict";
+  let hasCanonicalizer = 1;
 }
 
-def PCL_PostOp : PCL_Op<"post-cond", []> {
+def PCL_PostOp : PCL_Op<"post_cond", []> {
```
shankarapailoor, line 193: Q: Why change it? Is it more conventional?

backends/pcl/lib/Dialect/Ops.cpp
```diff
@@ -10,7 +10,581 @@
...
+// Ops over fields (In alphabetical order)
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+// AddOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AddOp::fold(FoldAdaptor adaptor) {
+  return tryFoldBinaryFeltOp(*this, adaptor, [](auto lhs, auto rhs) {
+    return FeltAttr::get(lhs.getContext(), lhs.getValue() + rhs.getValue());
```
shankarapailoor: Shouldn't you first zero extend both `lhs` and `rhs` to `max(bitwidth(lhs), bitdwidth(rhs)) + 1`?

backends/pcl/lib/Dialect/Ops.cpp
```diff
@@ -10,7 +10,581 @@
...
+//===----------------------------------------------------------------------===//
+
+OpFoldResult ConstOp::fold(FoldAdaptor adaptor) { return adaptor.getValue(); }
+
+//===----------------------------------------------------------------------===//
+// MulOp
+//===----------------------------------------------------------------------===//
+
+namespace {
+struct FoldXTimesMinus1 : public OpRewritePattern<MulOp> {
```
shankarapailoor: I might rename this to Fold0Minus1ToNeg1 or something. We should probably add a comment like:

// x * (p - 1) -> -x

because we mean `p - 1` for -1

backends/pcl/lib/Dialect/Ops.cpp
```diff
@@ -10,7 +10,581 @@
...
+
+} // namespace
+
+void SubOp::getCanonicalizationPatterns(RewritePatternSet &patterns, MLIRContext *context) {
+  patterns.add<ZeroMinusXToNegX>(context);
+}
+
+OpFoldResult SubOp::fold(FoldAdaptor adaptor) {
+  return tryFoldBinaryFeltOp(*this, adaptor, [](auto lhs, auto rhs) {
+    return FeltAttr::get(lhs.getContext(), lhs.getValue() - rhs.getValue());
```
shankarapailoor: We need to be careful here because it could underflow. I might rewrite `l - r` to `l + (p - (r % p))` and make sure you do the right zero extensions

backends/pcl/lib/Conversion/PCLLoweringPass.cpp
```diff
@@ -802,12 +802,12 @@ class PassImpl : public pcl::impl::PCLLoweringPassBase<PassImpl> {
 
   // PCL programs require a module-level attribute specifying the prime.
   LogicalResult setPrime(llvm::APInt &prime) {
-    // Add an extra bit to avoid the prime being represented as a negative number
-    auto newBitWidth = prime.getBitWidth() + 1;
-    auto ty = IntegerType::get(&getContext(), newBitWidth);
-    auto intAttr = IntegerAttr::get(ty, prime.zext(newBitWidth));
+    // // Add an extra bit to avoid the prime being represented as a negative number
```
shankarapailoor: Let's remove the commented code.

backends/pcl/lib/Dialect/Ops.cpp
```diff
@@ -10,7 +10,581 @@
...
+    return matchOperandsImpl(op.getRhs(), op.getLhs(), prime);
+  }
+
+  FailureOr<Value> matchOperandsImpl(Value lhs, Value rhs, PrimeAttr prime) const {
+    auto feltAttr = getAttr(lhs);
+    if (!feltAttr) {
+      return failure();
+    }
+    auto diff = prime.getValue() - feltAttr.getValue();
+    llvm::dbgs() << "Prime: " << prime.getValue() << "\nValue: " << feltAttr.getValue()
```
shankarapailoor: Should we use `LLVM_DEBUG(..)`

backends/pcl/lib/Dialect/Ops.cpp
```diff
@@ -10,7 +10,581 @@
...
+  patterns.add<FoldEqBoolean>(context);
+}
+
+//===----------------------------------------------------------------------===//
+// CmpLtOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult CmpLtOp::fold(FoldAdaptor adaptor) {
+  return foldCmpOp(*this, adaptor, [](auto lhs, auto rhs) {
+    return lhs.getValue().ult(rhs.getValue());
```
shankarapailoor: We have to be careful with the comparisons because you don't guarantee `lhs` and `rhs` are in the field. Same issue with less than or equal, greater than or equal, etc...

backends/pcl/lib/Dialect/Ops.cpp
```diff
@@ -10,7 +10,581 @@
...
+
+//===----------------------------------------------------------------------===//
+// NotOp
+//===----------------------------------------------------------------------===//
+
+namespace {
+struct FoldDoubleNeg : public OpRewritePattern<NotOp> {
+  using OpRewritePattern<NotOp>::OpRewritePattern;
+
+  LogicalResult match(NotOp op) const override {
```
shankarapailoor: Can we implement the `matchAndRewrite` function and merge the `match` and `rewrite` into one? I think they are deprecated in newer versions no?

shankarapailoor (changes requested):
I know this is a draft but pointing out bugs, some of which may fail silently

backends/pcl/include/pcl/Dialect/IR/Ops.td
0xddom: I was getting parse errors with the `-`, so I guess that name was not a valid op mnemonic

backends/pcl/lib/Dialect/Ops.cpp
0xddom: Yeah, that seems to be the case

backends/pcl/lib/Dialect/Ops.cpp
0xddom: I changed the helper to ensure that the value passed to the callback is within the field

backends/pcl/lib/Dialect/Ops.cpp
```diff
@@ -10,7 +10,619 @@
...
+) {
+  auto prime = getFieldPrime(op);
+  if (!prime) {
+    return nullptr;
+  }
+
+  return foldBinaryOp<FeltAttr>(op, adaptor, [&prime, opFn](FeltAttr lhs, FeltAttr rhs) {
+    lhs = prime->reduce(lhs);
+    rhs = prime->reduce(rhs);
+    auto max = std::max(
```
shankarapailoor: For multiplication you need to zero extend to `w_lhs+w_rhs` (probably `w_lhs + w_rhs + 1` to be safe). The rest look good.

backends/pcl/lib/Dialect/Dialect.cpp
```diff
@@ -54,10 +63,49 @@ PCLDialect::verifyOperationAttribute(mlir::Operation *op, mlir::NamedAttribute a
...
+FeltAttr PrimeAttr::reduce(FeltAttr attr) {
+  auto max = std::max({getValue().getBitWidth(), attr.getValue().getBitWidth()}) + 1;
+  auto pExt = getValue().zext(max);
+  // The incoming value could be negative so we need to sign-extend.
+  auto vExt = attr.getValue().sext(max);
+  auto value = vExt.srem(pExt);
+  if (value.isNegative()) {
+    value += pExt;
+  }
+  return FeltAttr::get(getContext(), value);
```
shankarapailoor: `reduce` is close but you need to be careful because different values being reduced will have different widths. This can bork things like comparisons. 

Say you are doing 4 < 15 over $$F_7$$. Then $$15 \mod 7$$ will have a width of 9 and $$4 \mod 7$$ will have a bitwidth of 5 and the comparison will complain. 

We should make the output have the same width as the modulus.

backends/pcl/lib/Dialect/Dialect.cpp
```diff
@@ -54,10 +63,49 @@ PCLDialect::verifyOperationAttribute(mlir::Operation *op, mlir::NamedAttribute a
...
+    return nullptr;
+  });
+}
+
+//===----------------------------------------------------------------------===//
+// PrimeAttr
+//===----------------------------------------------------------------------===//
+
+namespace {
+
```
shankarapailoor: NIT: We can delete this right?

shankarapailoor (changes requested):
A few more fixes needed but it is in better shape. 

I would make sure to add folding tests for comparison ops because there are a few width issues that would be caught with some basic tests.

backends/pcl/lib/Dialect/Dialect.cpp
0xddom: > We should make the output have the same width as the modulus.

That's a good point

backends/pcl/lib/Dialect/Ops.cpp
0xddom: Doesn't `APInt` handle the extension automatically?

0xddom:
@shankarapailoor re-requesting a review since I added the changes I wanted to make to the PCL->Lisp translation. I've not made the changes to the APInt related stuff you mentioned in the last pass.

There are also some compilation errors that for some reason didn't happen on my machine.

shankarapailoor:
OK sounds good. Yikes idk why I didn't use type switches.

backends/pcl/lib/Dialect/Ops.cpp
shankarapailoor: I don't think so. I think it will complain if the bitwidths aren't the same: https://github.com/llvm/llvm-project/blob/22116c946aa8078cee8d0c5efa49874c6264e92e/llvm/lib/Support/APInt.cpp#L236

backends/pcl/lib/Target/PCL.cpp
```diff
@@ -32,155 +40,269 @@ struct NameState {
...
+    }
+    return llvm::TypeSwitch<Operation *, FailureOr<pcl::Sexp>>(op)
+        .Case<pcl::AddOp>([this, &S](auto op) { return emitBinaryExpr("+", op, S); })
+        .Case<pcl::MulOp>([this, &S](auto op) { return emitBinaryExpr("*", op, S); })
+        .Case<pcl::SubOp>([this, &S](auto op) { return emitBinaryExpr("-", op, S); })
+        .Case<pcl::NegOp>([this, &S](auto op) { return emitUnaryExpr("-", op.getVal(), S); })
+        .Case<pcl::AsFeltOp>([this, &S](auto op) { return emitFormula(op.getVal(), S); })
+        .Case<pcl::VarOp>([this, &S](auto op) { return S.atom(ns.get(op)); })
+        .Case<pcl::ConstOp>([&S](auto op) {
+      return S.atom(op.getValue());
```
shankarapailoor: Issue: I think this will emit the raw felt attribute instead of the integer. It should be `S.atom(op.getValueAPInt())`.

backends/pcl/include/pcl/Dialect/IR/Ops.td
```diff
@@ -97,67 +104,95 @@ def PCL_AsFeltOp : PCL_Op<"asfelt", [Pure, InsideFunc]> {
...
+def PCL_CmpEqOp : PCLCmpBase<"eq", "lhs == rhs (field equality)", true>;
 def PCL_CmpLtOp : PCLCmpBase<"lt", "lhs < rhs (signed/semantic: range)">;
 def PCL_CmpLeOp : PCLCmpBase<"le", "lhs <= rhs">;
 def PCL_CmpGtOp : PCLCmpBase<"gt", "lhs > rhs">;
 def PCL_CmpGeOp : PCLCmpBase<"ge", "lhs >= rhs">;
 
+// -----------------------------------------------------------------------------
+// Boolean constants
+// -----------------------------------------------------------------------------
+def PCL_TrueOp : PCL_Op<"true", [Pure, InsideFunc, ConstantLike]> {
```
shankarapailoor: Can you remind me why you are adding true and false ops? It seems they lower to 1 and 0.

backends/pcl/lib/Target/PCL.cpp
```diff
@@ -32,155 +40,269 @@ struct NameState {
...
+  pcl::Sexps S;
+
+  if (failed(prologue(mod, S))) {
+    return failure();
   }
+
+  S.dump(os);
+  nl(os);
+
+  for (auto emitter : findModules(mod)) {
```
shankarapailoor:
```suggestion
  for (auto &emitter : findModules(mod)) {
```

backends/pcl/lib/Dialect/Dialect.cpp
```diff
@@ -12,8 +12,14 @@
 #include "pcl/Dialect/IR/Ops.h"
 #include "pcl/Dialect/IR/Types.h"
 
+#include "llzk/Dialect/LLZK/IR/AttributeHelper.h"
+
 #include <mlir/IR/DialectImplementation.h>
 
+#include <llvm/Support/Debug.h>
```
shankarapailoor: Is this used?

backends/pcl/lib/Target/Sexp.cpp
```diff
@@ -0,0 +1,50 @@
...
+  elt->setParen('[');
+  return *this;
+}
+
+Sexp SexpCtx::sexp(llvm::ArrayRef<Sexp> elements) {
+  detail::SexpElt **buf = allocator.Allocate<detail::SexpElt *>(elements.size());
+  memcpy(
+      reinterpret_cast<void *>(buf), elements.data(), elements.size() * sizeof(detail::SexpElt *)
+  );
+  return Sexp(new (allocator) detail::List(llvm::ArrayRef(buf, elements.size())));
```
shankarapailoor, line 47: NIT: maybe add `static_assert(sizeof(Sexp) == sizeof(detail::SexpElt *))`?

backends/pcl/include/pcl/Dialect/IR/Ops.td
0xddom: For having a constant boolean when folding ops like `and` or `or`. The const op returns a felt type so I need something that returns a bool type.

backends/pcl/lib/Target/Sexp.cpp
0xddom: Yeah, that's the intention so should be a good idea for peace of mind

backends/pcl/lib/Dialect/Dialect.cpp
0xddom: No, it was a left over of some debugging session

backends/pcl/lib/Target/PCL.cpp
0xddom: Yeah, good catch

backends/pcl/include/pcl/Dialect/IR/Attrs.td
```diff
@@ -12,21 +12,26 @@
...
+  // let builders = [AttrBuilder<(ins "const ::llvm::APInt &":$v), [{
+  //     return $_get(context,
+  //     ::mlir::IntegerAttr::get(::mlir::IntegerType::get(context,
+  //     v.getBitWidth()), v));
+  //   }]>,
+  //                 AttrBuilder<(ins "const ::llvm::APSInt &":$v), [{
+  //     return $_get(context,
+  //     ::mlir::IntegerAttr::get(::mlir::IntegerType::get(context,
+  //     v.getBitWidth()), v));
+  //   }]>];
```
tim-hoffman: use it or lose it

backends/pcl/include/pcl/Dialect/IR/Attrs.td
```diff
@@ -36,8 +41,27 @@ def PrimeAttr : AttrDef<PCLDialect, "Prime"> {
...
+  ::pcl::FeltAttr reduce(::pcl::FeltAttr attr);
+  ::pcl::FeltAttr reduce(::llvm::APInt value) {
+    return reduce(::pcl::FeltAttr::get(getContext(), value));
+  }
+  }];
+}
+
+def PCLBoolAttr : AttrDef<PCLDialect, "Bool"> {
+  let mnemonic = "bool";
+  let cppClassName = "BoolAttr";
```
tim-hoffman: This is redundant. Similar in existing `FeltAttr` above.

backends/pcl/include/pcl/Dialect/IR/Ops.td
tim-hoffman: I think that hypen had cause me an annoyance in the CAPI stuff as well 😆

backends/pcl/include/pcl/Dialect/IR/Ops.td
```diff
@@ -97,67 +105,95 @@ def PCL_AsFeltOp : PCL_Op<"asfelt", [Pure, InsideFunc]> {
...
+  let summary = "constant false value";
+  let results = (outs PCL_BoolType:$res);
+  let assemblyFormat = "attr-dict";
+  let hasFolder = 1;
+}
+
 // -----------------------------------------------------------------------------
 // Boolean connectives on !pcl.bool
 // -----------------------------------------------------------------------------
 def PCL_AndOp : PCL_Op<"and", [Pure, Commutative, SameOperandsAndResultType]> {
```
tim-hoffman: could simplify a lot of these by moving common stuff to `binop` and `unop` classes

backends/pcl/lib/Dialect/Dialect.cpp
```diff
@@ -54,10 +62,63 @@ PCLDialect::verifyOperationAttribute(mlir::Operation *op, mlir::NamedAttribute a
...
 }
+
+mlir::Operation *PCLDialect::materializeConstant(
+    mlir::OpBuilder &builder, mlir::Attribute value, mlir::Type, mlir::Location loc
+) {
+  return llvm::TypeSwitch<mlir::Attribute, mlir::Operation *>(value)
+      .Case<FeltAttr>([&builder, loc](auto attr) -> mlir::Operation * {
+    return builder.create<ConstOp>(loc, attr);
+  })
+      .Case<BoolAttr>([&builder, loc](auto attr) -> mlir::Operation * {
```
tim-hoffman, line 80: You may not need the return type since it's specified on the TypeSwitch itself
```suggestion
      .Case<FeltAttr>([&builder, loc](auto attr) {
    return builder.create<ConstOp>(loc, attr);
  })
      .Case<BoolAttr>([&builder, loc](auto attr) {
```

test/PCL/canonicalize_pass.llzk
```diff
@@ -0,0 +1,1004 @@
...
+// CHECK-NEXT:    func.func @and() -> !pcl.bool {
+// CHECK-NEXT:      %[[VAL_0:[0-9a-zA-Z_\.]+]] = pcl.true
+// CHECK-NEXT:      return %[[VAL_0]] : !pcl.bool
+// CHECK-NEXT:    }
+// CHECK-NEXT:  }
+
+// -----
+
+module attributes {pcl.prime = #pcl.prime<7>} {
+  func.func @and() -> !pcl.bool {
```
tim-hoffman: Can you rename all the `@and`/`@or` tests to make it obvious which folding scenario each is testing.

test/PCL/canonicalize_pass.llzk
```diff
@@ -0,0 +1,1004 @@
...
+// CHECK-NEXT:    }
+// CHECK-NEXT:  }
+
+// -----
+
+module attributes {pcl.prime = #pcl.prime<7>} {
+  func.func @implies() -> !pcl.bool {
+    %0 = pcl.true
+    %1 = pcl.false
+    %2 = pcl.implies %0, %1
```
tim-hoffman, line 753: What about tests for the other direction of implies?

test/PCL/canonicalize_pass.llzk
```diff
@@ -0,0 +1,1004 @@
...
+// CHECK-NEXT:    }
+// CHECK-NEXT:  }
+
+// -----
+
+module attributes {pcl.prime = #pcl.prime<7>} {
+  func.func @iff() -> !pcl.bool {
+    %0 = pcl.true
+    %1 = pcl.false
+    %2 = pcl.iff %0, %1
```
tim-hoffman, line 861: Test for the other direction as well since it's commutative

backends/pcl/lib/Target/PCL.cpp
```diff
@@ -1,26 +1,35 @@
...
+#include <llvm/ADT/SmallVectorExtras.h>
+#include <llvm/ADT/StringRef.h>
+#include <llvm/ADT/TypeSwitch.h>
+#include <llvm/Support/Allocator.h>
+#include <llvm/Support/raw_ostream.h>
 
+using namespace mlir;
 namespace {
+
 struct NameState {
```
tim-hoffman, line 34: can you add documentation on this

backends/pcl/include/pcl/Dialect/IR/Attrs.td
0xddom: Completely forgot about it

backends/pcl/lib/Dialect/Dialect.cpp
0xddom: Nope, it fails to build. I think its because the lambda's inferred return type is one of the "named" ops instead of the generic pointer to operation.

test/PCL/canonicalize_pass.llzk
```diff
@@ -1002,3 +1145,178 @@ module attributes {pcl.prime = #pcl.prime<7>} {
...
+
+// CHECK-LABEL: module attributes {pcl.prime = #pcl.prime<7>} {
+// CHECK-NEXT:    func.func @fold_eq_rhs_0_and_sum_lhs_neg() {
+// CHECK-NEXT:      %[[VAL_0:[0-9a-zA-Z_\.]+]] = pcl.var "a" false
+// CHECK-NEXT:      %[[VAL_1:[0-9a-zA-Z_\.]+]] = pcl.var "b" false
+// CHECK-NEXT:      %[[VAL_2:[0-9a-zA-Z_\.]+]] = pcl.eq %[[VAL_0]], %[[VAL_1]]
+// CHECK-NEXT:      pcl.assert %[[VAL_2]]
+// CHECK-NEXT:      return
+// CHECK-NEXT:    }
+// CHECK-NEXT:  }
```
tim-hoffman, line 1321: Is there folding for `pcl.sub` similar to these last test cases for add+neg?

tim-hoffman (approved):
LGTM; just one question about whether more tests should be added for `pcl.sub`

test/PCL/canonicalize_pass.llzk
0xddom: Ah, yes. `(= 0 (- x y))` folds to `(= x y)`. I thought I added the tests for those

shankarapailoor (approved):
LGTM

PR #628, opened 2026-07-22
produce an error diagnostic instead of assertion failure

tim-hoffman:
The new test case previously failed the `assert(productFn)`. It must produce an error diagnostic instead.


shankarapailoor (approved):
LGTM

PR #630, opened 2026-07-22
Add assert/prove ops to verif and SMT

raghav198:


include/llzk/Dialect/Verif/IR/Ops.td
```diff
@@ -46,12 +44,31 @@ class ConditionOp<string mnemonic, list<Trait> traits = []>
...
+  let summary = "An assertion in the target of a contract, derived from one of "
+                "the contract's specifications";
+  let description = summary;
+}
+
+def VerifProveOp : ConditionOp<"prove", []> {
+  let summary = "A proof obligation in the target of a contract, derived from"
+                "one of the contract's specifications";
+  let description = summary;
+}
```
iangneal, line 65: These seem eerily similar to `require_*` and `ensure_*` ops. I'm sure you have some intended difference but from this description alone it is not obvious to me what purpose these serve.

include/llzk/Dialect/SMT/IR/SMTOps.td
```diff
@@ -176,6 +176,12 @@ def AssertOp : SMTOp<"assert", []> {
   let assemblyFormat = "$input attr-dict";
 }
 
+def ProveOp : SMTOp<"prove", []> {
```
iangneal: Does this have a direct translation to SMTLIB2 expressions or not? Most ops here do, so if this doesn't, it'd be helpful to note in the description.

include/llzk/Dialect/Verif/IR/Ops.td
raghav198: Oh, yeah. The difference here is in applying the contract to a particular target, e.g. a `require` precondition has different semantics depending on whether its being applied to the main component (becomes a hypothesis) or a subcomponent (becomes a proof obligation), whereas an `ensure` postcondition is kind of dual (its a proof obligation in the main component, but turns into a hypothesis for a subcomponent). I tried to clarify the distinction by saying these ops show up in the contract's _target_ and not in the original contract; the idea is to eventually write some kind of lowering that applies a contract to a target (possibly with subcomponent calls) by translating all the relevant requires/ensures into these more specific ops over concrete values.

changelogs/unreleased/raghav__verif-ops.yaml
```diff
@@ -0,0 +1,3 @@
+added:
+  - Ops `verif.assert` and `verif.prove` to `verif` dialect
+  - Op `smt.prove` to `smt` dialect
```
tim-hoffman: update

include/llzk/Dialect/Verif/IR/Ops.td
```diff
@@ -46,12 +45,38 @@ class ConditionOp<string mnemonic, list<Trait> traits = []>
...
+
+def VerifSMTProveOp : Op<VerifDialect, "smt_prove", []> {
+  let arguments = (ins BoolType:$condition);
+  let results = (outs);
```
tim-hoffman: Empty is redundant
```suggestion
```

test/Dialect/Verif/target_pass.llzk
```diff
@@ -0,0 +1,79 @@
...
+module attributes {llzk.lang} {
+    struct.def @A {
+        function.def @product(%x: !felt.type, %y: !felt.type) -> !struct.type<@A> {
+            %self = struct.new : <@A>
+            %ok = bool.cmp lt(%x, %y)
+            verif.assert %ok
+            function.return %self : !struct.type<@A>
+        }
+    }
+
```
tim-hoffman: nit: remove empty line

include/llzk/Dialect/Verif/IR/Ops.td
iangneal: > a `require` precondition has different semantics depending on whether its being applied to the main component (becomes a hypothesis) or a subcomponent (becomes a proof obligation)

Seems like `assert` would have the same set of semantics (i.e., `assert` in main is a pre-condition, an an `assert` in a subcomponent is a proof obligation for the parent component).

> whereas an ensure postcondition is kind of dual (its a proof obligation in the main component, but turns into a hypothesis for a subcomponent).

I'm not sure an `ensure` post-condition in a parent component becomes a hypothesis for a subcomponent, it seems like parent post-conditions are not taken for granted in subcomponents.

Either way, it still seems like `ensure_*` are proof obligations and `require_*` is either a hypothesis for the entry struct or a proof obligation when in a subcomponent. What if instead of introducing these new ops, we:

- added unqualified `verif.require` and `verif.ensure` that enforced both `verif.require_compute/constrain` and `verif.ensure_compute/constrain`, respectively
- Loosened the restrictions on require/ensure ops to allow them within targets directly, instead of only within contracts

It seems like this would accomplish the same goal without introducing a new set of semantics.

raghav198: Ah, yeah sorry I don't think I explained correctly. Here's the use case I have in mind:

```
struct @Sub {
    ...
}
struct @Top {
    function.def @product(%top_inputs) {
        ...
        %sub = @Sub::@product(%sub_inputs);
        ...
    }
}
```
with contracts

```
verif.contract for @Sub {
    require P
    ensure Q
}
verif.contract for @Top {
    require R
    ensure S
}
```

A lowering pass that applies the contract to `@Top` should give you something like:

```
struct @Top {
    function.def @product(%top_inputs) {
        verif.assert R(%top_inputs) // take the precondition of @Top as a hypothesis
        ...
        verif.prove P(%sub_inputs) // prove that the call to @Sub satisfies @Sub's preconditions
        %sub = @Sub::@product(%sub_inputs);
        verif.assert Q // since we've proven that @Sub's preconditions are satisfied, take @Sub's postconditions as a hypothesis
        verif.prove S // now prove @Top's postconditions
        ...
    }
}
```
so the final verification condition for this looks something like `R(top) /\ Q ==> P(sub) /\ S`

In particular, `require P` and`ensure S` both turn into proof obligations, whereas `ensure Q` and `require R` both turn into assertions/hypothesis.

shankarapailoor: I'm a bit confused by the final verification condition you wrote: `R(top) /\ P(sub) ==> Q /\ S`. Why would a proof obligation $$P$$ be in the antecedent for the final verification condition?

raghav198: Oops, because I made a typo :( Fixed it now

iangneal: Okay, I think I understand the difference between these things now. @raghav198 correct me if I'm wrong, but I think we're saying:
- `verif.require_*` declares a target's preconditions
- `verif.ensure_*` declares a target's postconditions
- `verif.assert` is an imperative statement saying to add the given condition to the running solver.
- `verif.prove` is an imperative statement telling the solver to prove the given condition given the current set of assertions

So in effect the `verif.assert` and `verif.prove` are more "proving runtime" operations than declarations of semantics. Yes?

iangneal: In this way `verif.assert` is akin to `smt.assert` and `verif.prove` is like `smt.check`

raghav198: > So in effect the verif.assert and verif.prove are more "proving runtime" operations than declarations of semantics. Yes?

Yeah

> In this way verif.assert is akin to smt.assert and verif.prove is like smt.check

Kind of, but `smt.check` I think directly just checks whether all the assertions in the context, taken together _can be true_, whereas `verif.prove` is asking whether the given expression _must be true_ given all the assertions in the context.

Like it can't literally lower to `smt.check` but it is morally similar, yeah.

raghav198: In particular, you could do something like lowering `verif.prove %b` to `smt.assert (not %b); smt.check` and then look for an UNSAT result, but this is kind of forcing a bunch of implementation decisions that the verifier might be able to do better with.

iangneal: @raghav198 in our discussion yesterday you mentioned potentially separating `verif.assert` and `verif.prove` by compute/constrain, what are your current thoughts on that?

raghav198: I was planning to just use the `product_source` attribute like the product alignment passes currently do (although at this point it would maybe be better to rename the attribute to something more generic) since SMT lowering already has to check that when deciding whether to target `_c` or `_w`.

iangneal: The renaming sounds like a good idea, but probably not for this PR, maybe make an issue for it.

tim-hoffman (approved):
LGTM

PR #604, opened 2026-07-14
Support felt constant template parameters in witness generation

1sgtpepper:
Closes #479.

## Summary

Circom-produced felt constants can now flow through template flattening and witness
generation as concrete template arguments. Adding the Circom witness examples exposed
that flattening did not yet preserve the felt and partial-specialization semantics needed
by those definitions.

- Treats felt constants as concrete template arguments during flattening.
- Makes fieldless constants adopt the required felt type and rejects incompatible
  explicitly fielded constants with source/expected-field diagnostics.
- Evaluates template expressions when their dependencies are concrete and preserves
  deferred expressions, including rewritten concrete dependencies and dependencies in
  nested regions, through partial struct and free-function specialization.
- Reuses specializations by source definition and structural parameter bindings rather
  than generated symbol spelling; generated names remain collision-safe and escape
  reserved bytes.
- Adds Circom witness-generation coverage for both interpreter and execution-engine
  backends, plus transform and unit coverage for field compatibility, partial
  specialization, cache identity, and symbol collisions.

Integer and type template parameters retain their existing behavior. This change does
not expand template-expression evaluation to region-bearing operations.

## Validation

- FileCheck coverage generated with the repository's `scripts/generate-test-checks.py` utility.
- GitHub Actions at `030055e`: all 9 checks passed, including GCC, Clang,
  debug/release Nix builds, formatting, docs, changelog, and aggregate results.

PR #633, opened 2026-07-23
Make specialization name encoding collision-safe

1sgtpepper:
Part of #479 (approved); split from #604.

## Summary

Preserve literal source-name bytes and partial-template name positions across specialization passes. Partial function reuse is keyed by the exact source function and concrete bindings rather than rendered symbol names.

## Changes

- Add verified `poly.name_pattern` metadata containing literal chunks around the current template parameters; absent metadata leaves the entire symbol name opaque.
- Preserve post-insertion template and function names when symbol-table uniquing changes a preferred generated name.
- Make felt specialization names distinguish field names with a byte-length prefix.
- Add positive, negative, collision, cache-reuse, wildcard, and type-inference coverage.

## Testing

- GitHub Actions: all 9 checks pass, including GCC, Clang, debug/release, docs, formatting, changelog, and test-result checks.
- Focused `llzk-opt`/FileCheck and `-verify-diagnostics` pipelines for flattening, type-variable inference, wildcard specialization, malformed `poly.name_pattern`, and the existing felt-field diagnostic.

## Documentation and changelog

The `poly.template` TableGen documentation describes the metadata contract, and the required unreleased changelog entry is included.

Allow edits from maintainers is enabled.

## AI assistance

- [ ] No AI tools contributed to this PR.
- [x] AI tools contributed to this PR.

**Tools used:** Codex

**How the tools contributed:** Implementation.

**How I verified the contribution:** Reviewed the final diff and confirmed all 9 GitHub Actions checks passed.


tim-hoffman:
@codex review

iangneal:
@codex review

lib/Util/TypeHelper.cpp
```diff
@@ -44,6 +44,22 @@ using namespace polymorphic;
 using namespace string;
 using namespace pod;
 
+std::string BuildShortTypeString::escapeSpecialCharacters(StringRef value) {
```
iangneal: This is a duplicate of some code you have in another PR, so refer to my comments there: https://github.com/project-llzk/llzk-lib/pull/634/changes#r3646683046

test/Transforms/Flattening/instantiate_name_encoding.llzk
```diff
@@ -0,0 +1,93 @@
...
+// CHECK-NEXT: function.def @compute() -> index {
+// CHECK-NEXT: %[[VAL_2:[0-9a-zA-Z_\.]+]] = poly.read_const @N : index
+// CHECK-NEXT: function.return %[[VAL_2]] : index
+// CHECK-NEXT: }
+// CHECK-NEXT: }
+// CHECK-NEXT: function.def @main() -> index {
+// CHECK-NEXT: %[[VAL_3:[0-9a-zA-Z_\.]+]] = function.call @"T%[[VAL_4:[0-9a-zA-Z_\.]+]]A%[[VAL_4]]A_35_compute"() : () -> index
+// CHECK-NEXT: function.return %[[VAL_3]] : index
+// CHECK-NEXT: }
+// CHECK-NEXT: }
```
iangneal: Nit but can you indent this to match the source indent style? Makes it easier to read.

test/Transforms/Flattening/instantiate_name_encoding.llzk
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  function.def @main() -> index {
+    %got = function.call @"T\1A\1A"::@compute<[35]>() : () -> index
+    function.return %got : index
+  }
+}
+
+// CHECK-LABEL: module attributes {llzk.lang} {
+// CHECK-NEXT: function.def @"T%[[VAL_0:[0-9a-zA-Z_\.]+]]A%[[VAL_0]]A_35_compute"() -> index {
```
iangneal: I think I'd actually prefer to see the raw output instead of the `VAL_0` capture and reuse. Same below.

tim-hoffman:
@1sgtpepper there's a real bug here as witnessed by `test/Transforms/Flattening/instantiate_name_encoding.llzk` but I think this solution is a bit too adhoc and can result in some problems.

Multiple passes can perform some kind of partial instantiation that adds the placeholder character: `specialize-wildcard-arrays`, `inline-structs`, `infer-tvar`, and of course `flatten` (where it's expected that placeholders would be replaced). The problem with `fromRawName` and the other uses of `escapeSpecialCharacters` is that they may be called when a string is not actually a "raw name" because an earlier pass created this name from `BuildShortTypeString` but the later pass has no way of knowing that. So, a placeholder added by `infer-tvar` can be escaped during `flatten` when it was expected that `flatten` would now replace it with the proper instantiation.

One foolproof solution would be a verification on all input IR to ensure the `PLACEHOLDER` is not present.

tim-hoffman (changes requested):
see earlier comment

iangneal:
> @1sgtpepper there's a real bug here as witnessed by `test/Transforms/Flattening/instantiate_name_encoding.llzk` but I think this solution is a bit too adhoc and can result in some problems.
> 
> Multiple passes can perform some kind of partial instantiation that adds the placeholder character: `specialize-wildcard-arrays`, `inline-structs`, `infer-tvar`, and of course `flatten` (where it's expected that placeholders would be replaced). The problem with `fromRawName` and the other uses of `escapeSpecialCharacters` is that they may be called when a string is not actually a "raw name" because an earlier pass created this name from `BuildShortTypeString` but the later pass has no way of knowing that. So, a placeholder added by `infer-tvar` can be escaped during `flatten` when it was expected that `flatten` would now replace it with the proper instantiation.
> 
> One foolproof solution would be a verification on all input IR to ensure the `PLACEHOLDER` is not present.

This feedback also applies to #634 which has a similar function.

1sgtpepper:
Should `PLACEHOLDER` be rejected before transforms run, including in included modules?

iangneal:
> Should `PLACEHOLDER` be rejected before transforms run, including in included modules?

Probably so, but we'll get @tim-hoffman's input as well.

tim-hoffman:
> > Should `PLACEHOLDER` be rejected before transforms run, including in included modules?
> 
> Probably so, but we'll get @tim-hoffman's input as well.

I'm putting together some test cases to exercise the specific problems I mentioned.

1sgtpepper:
Ill hold off on #634 until this is cleared  so I can reconcile the two.

tim-hoffman:
@1sgtpepper https://github.com/project-llzk/llzk-lib/pull/646 adds some tests that would end up with mangled/reordered names due to changes in this PR. Hopefully that clears up the issue I was describing.

In the "foolproof" solution I mentioned before, modules included via `IncludeOp::inlineAndErase()` would also need to be sanitized. However, I hadn't really considered protecting the surface of IR generated via the c/c++ "build" API functions from applications using this repo as a library. That's more challenging. I am open to any ideas you have on how to handle this, even so far as replacing the `PLACEHOLDER` based implementation with some other way to track positions in symbol names that will be replaced with attribute values in the future.

1sgtpepper:
Thanks for #646 been looking into this and seems more messy than I initially thought, I'll share my thoughts soon since this also affects my adjacent prs substantially(it seems)

1sgtpepper:
Thanks—#646 convinced me that the spelling itself is the wrong place to retain slot state.

My proposal is to store literal name chunks on `poly.template`, with gaps matching the remaining parameters. Passes refine those gaps; `sym_name` remains opaque.

Two questions:

1. Should this be an inherent ODS attribute, or is a verified generic `ArrayAttr` appropriate?
2. Can absence mean a literal source/legacy name, with no automatic migration?

I'd keep the rest out of this PR.

1sgtpepper:
Leaning toward an optional inherent ODS name_pattern backed by ArrayAttr, since later passes rely on it.

When absent, the entire sym_name is opaque, regardless of whether parameters remain; slots are never inferred from marker bytes.

I’d keep the scope to the existing partial-template refinement paths, plus a pass-local exact cache replacing rendered-name reuse for partial functions.

tim-hoffman:
@1sgtpepper I'd lean toward a verified generic `ArrayAttr` rather than adding it to the ODS since it's not truly a language feature but rather an artifact needed only for transformation passes. It's unclear to me how this would work when multiple versions of the same template are created; is the symbol name also being updated to avoid conflicts?

1sgtpepper:
A verified generic `ArrayAttr` works. Each generated partial-template clone gets a rendered `sym_name`; `SymbolTable` uniquifies collisions, and references use the inserted name. Flattening reuses partial functions by source function and exact bindings.

Is it okay for a missing attribute to mean the entire name is opaque, with no legacy slot reconstruction?

tim-hoffman:
> A verified generic `ArrayAttr` works. Each generated partial-template clone gets a rendered `sym_name`; `SymbolTable` uniquifies collisions, and references use the inserted name. Flattening reuses partial functions by source function and exact bindings.
> 
> Is it okay for a missing attribute to mean the entire name is opaque, with no legacy slot reconstruction?

Yes. That should be fine.

test/Transforms/Flattening/instantiate_name_encoding.llzk
```diff
@@ -0,0 +1,328 @@
...
+// CHECK-NEXT:    function.def @main() -> index {
+// CHECK-NEXT:      %[[VAL_2:[0-9a-zA-Z_\.]+]] = function.call @T_89_13_compute() : () -> index
+// CHECK-NEXT:      function.return %[[VAL_2]] : index
+// CHECK-NEXT:    }
+// CHECK-NEXT:  }
+
+// -----
+
+module attributes {llzk.lang} {
+  // A previously partial template is refined from literal chunks, not from its physical name.
```
tim-hoffman:
```suggestion
  // A previously partial template is refined from `poly.name_pattern` chunks, not from its physical name.
```

test/Transforms/Flattening/instantiate_name_pattern_fail.llzk
```diff
@@ -0,0 +1,53 @@
+// RUN: llzk-opt -split-input-file -verify-diagnostics %s
+
+module attributes {llzk.lang} {
+  // The verifier owns the outer representation contract.
```
tim-hoffman: I don't understand what this comment is saying w.r.t. this test case.

test/Transforms/Flattening/instantiate_name_pattern_fail.llzk
```diff
@@ -0,0 +1,53 @@
...
+      %n = poly.read_const @N : index
+      function.return %n : index
+    }
+  } {poly.name_pattern = "T_"}
+}
+
+// -----
+
+module attributes {llzk.lang} {
+  // The verifier rejects non-literal chunks rather than assigning them slot meaning.
```
tim-hoffman, line 18:
```suggestion
  // The verifier rejects non-string chunks rather than assigning them slot meaning.
```

test/Transforms/Flattening/instantiate_name_pattern_fail.llzk
```diff
@@ -0,0 +1,53 @@
...
+      %n = poly.read_const @N : index
+      function.return %n : index
+    }
+  } {poly.name_pattern = ["T_", 42 : i64]}
+}
+
+// -----
+
+module attributes {llzk.lang} {
+  // expected-error@+1 {{expected 'poly.name_pattern' to contain 2 literal chunk(s), matching the template parameters, but found 1}}
```
tim-hoffman: "matching the template parameters" here is confusing since the number of chunks is greater than, not matching, the number of template parameters.

test/Transforms/Flattening/instantiate_name_pattern_fail.llzk
```diff
@@ -0,0 +1,53 @@
...
+      %n = poly.read_const @N : index
+      function.return %n : index
+    }
+  } {poly.name_pattern = ["T_"]}
+}
+
+// -----
+
+module attributes {llzk.lang} {
+  // expected-error@+1 {{expected 'poly.name_pattern' to contain 2 literal chunk(s), matching the template parameters, but found 3}}
```
tim-hoffman: Can you also add a case where it's expecting 1 chunk (or better yet, the `poly.name_pattern` attribute to not be used at all) because there are no template parameters?

lib/Util/TypeHelper.cpp
```diff
@@ -156,6 +156,14 @@ BuildShortTypeString &BuildShortTypeString::append(Attribute a) {
     Type ty = ia.getType();
     bool isUnsigned = ty.isUnsignedInteger() || ty.isSignlessInteger(1);
     ia.getValue().print(ss, !isUnsigned);
+  } else if (auto fa = llvm::dyn_cast<FeltConstAttr>(a)) {
+    ss << "f<";
+    fa.getValue().print(ss, false);
+    if (StringAttr fieldName = fa.getFieldName()) {
+      // The encoded byte length prevents field delimiters from colliding with adjacent parameters.
```
tim-hoffman: Is this what the `testShortStringDistinguishesDelimitedFeltFieldNames` test is covering? I think it would be helpful to put an explicit example in the comment here of what would be confused without adding the byte length.

include/llzk/Util/TypeHelper.h
```diff
@@ -34,8 +34,8 @@ class PodType;
 
 /// Note: If any symbol refs in an input Type/Attribute use any of the special characters that this
 /// class generates, they are not escaped. That means these string representations are not safe to
-/// reverse back into a Type. It's only intended to produce a unique name for instantiated structs
-/// that may give some hint when debugging regarding the original struct name and the params used.
+/// reverse back into a Type. They are intended only to produce compact, recognizable names for
+/// instantiated symbols.
 class BuildShortTypeString {
   static constexpr char PLACEHOLDER = '\x1A';
```
tim-hoffman: Can you explain if/why this one is still needed?

lib/Dialect/Polymorphic/IR/Ops.cpp
```diff
@@ -41,6 +41,38 @@ FailureOr<TemplateOp> verifyInTemplate(Operation *op) {
                            << "' ancestor";
 }
 
+/// Verify the optional transform-carried name pattern against current parameters.
+///
+/// The pattern is generic discardable metadata, so this owner boundary is the one place that
+/// prevents malformed cardinality or element types from reaching refinement passes.
+LogicalResult TemplateOp::verify() {
+  static constexpr llvm::StringLiteral namePatternAttr = "poly.name_pattern";
```
tim-hoffman: I'd rather not have this same string literal defined here and in `lib/Dialect/Polymorphic/Transforms/SharedImpl.h`. Please pick a reasonable shared header to define it once.

lib/Dialect/Polymorphic/Transforms/SharedImpl.h
```diff
@@ -131,52 +131,41 @@ inline component::StructType getStructTypeWithParams(
                         : component::StructType::get(nameRef, mlir::ArrayAttr::get(ctx, params));
 }
 
+/// Transform-private metadata carrying a partial template's logical name pattern.
+///
+/// When present, the attribute has P+1 StringAttr chunks around the P current template
+/// parameters. The chunks are literal data; later transforms never recover semantic positions
+/// from the physical symbol name.
```
tim-hoffman: Please clarify/expand this documentation. it's a little hard to follow.

lib/Dialect/Polymorphic/Transforms/SharedImpl.cpp
```diff
@@ -30,6 +30,150 @@
...
+  mlir::MLIRContext *ctx = parentTemplate.getContext();
+  mlir::SmallVector<mlir::Attribute> paramNames = parentTemplate.getConstNames<TemplateParamOp>();
+  mlir::SmallVector<mlir::StringAttr> sourceChunks;
+
+  if (mlir::Attribute rawPattern = parentTemplate->getDiscardableAttr(TEMPLATE_NAME_PATTERN_ATTR)) {
+    auto pattern = llvm::dyn_cast<mlir::ArrayAttr>(rawPattern);
+    if (!pattern) {
+      parentTemplate.emitOpError()
+          << "expected '" << TEMPLATE_NAME_PATTERN_ATTR << "' to be an ArrayAttr";
+      return mlir::failure();
```
tim-hoffman:
```suggestion
      return parentTemplate.emitOpError()
          << "expected '" << TEMPLATE_NAME_PATTERN_ATTR << "' to be an ArrayAttr";
```

lib/Dialect/Polymorphic/Transforms/SharedImpl.cpp
```diff
@@ -30,6 +30,150 @@
...
+    }
+    if (pattern.size() != paramNames.size() + 1) {
+      parentTemplate.emitOpError()
+          << "expected '" << TEMPLATE_NAME_PATTERN_ATTR << "' to contain " << paramNames.size() + 1
+          << " literal chunk(s), matching the template parameters, but "
+          << "found " << pattern.size();
+      return mlir::failure();
```
tim-hoffman:
```suggestion
      return parentTemplate.emitOpError()
          << "expected '" << TEMPLATE_NAME_PATTERN_ATTR << "' to contain " << paramNames.size() + 1
          << " literal chunk(s), matching the template parameters, but found " << pattern.size();
```

lib/Dialect/Polymorphic/Transforms/SharedImpl.cpp
```diff
@@ -30,6 +30,150 @@
...
+    }
+    for (auto [index, chunk] : llvm::enumerate(pattern)) {
+      auto stringChunk = llvm::dyn_cast<mlir::StringAttr>(chunk);
+      if (!stringChunk) {
+        parentTemplate.emitOpError() << "expected '" << TEMPLATE_NAME_PATTERN_ATTR << "' element "
+                                     << index << " to be a StringAttr";
+        return mlir::failure();
```
tim-hoffman:
```suggestion
        return parentTemplate.emitOpError() << "expected '" << TEMPLATE_NAME_PATTERN_ATTR 
                                     << "' element " << index << " to be a StringAttr";
```

lib/Dialect/Polymorphic/Transforms/SharedImpl.cpp
```diff
@@ -30,6 +30,150 @@
...
+  // Refine chunks in parameter order. A concrete gap is absorbed into the current literal chunk;
+  // an open gap starts the next chunk. This preserves empty chunks and therefore exact P+1 shape.
+  mlir::SmallVector<std::string> refinedChunkValues;
+  refinedChunkValues.push_back(sourceChunks.front().getValue().str());
+  for (size_t i = 0; i < paramNames.size(); ++i) {
+    auto concreteIt = paramNameToConcrete.find(paramNames[i]);
+    if (concreteIt != paramNameToConcrete.end() && concreteIt->second) {
+      mlir::Attribute binding = concreteIt->second;
+      mlir::SmallVector<mlir::Attribute, 1> oneBinding {binding};
+      refinedChunkValues.back() += BuildShortTypeString::from(oneBinding);
```
tim-hoffman: I think it's worth adding `BuildShortTypeString::from(Attribute)` in TypeHelper.h

unittests/IR/TypesTests.cpp
```diff
@@ -184,6 +196,66 @@ TEST_F(TypeTests, testShortString) {
...
+      ArrayAttr::get(&ctx, ArrayRef<Attribute> {felt(35, fieldA), unspecified})
+  );
+  std::string second = BuildShortTypeString::from(
+      ArrayAttr::get(&ctx, ArrayRef<Attribute> {felt(35, fieldB), felt(36, fieldC)})
+  );
+
+  EXPECT_NE(first, second);
+}
+
+TEST_F(TypeTests, testShortStringLengthPrefixesFeltFieldNames) {
```
tim-hoffman: Is this test irrelevant now without the need for escaping, etc.?

tim-hoffman (changes requested):

include/llzk/Util/TypeHelper.h
1sgtpepper: It is still needed by `BuildShortTypeString::from(base, attrs)` to preserve unresolved parameter positions across partial instantiations.

tim-hoffman: So, what scenarios are still managed via inserting `PLACEHOLDER` into the symbol name and later replacing it rather than using the `poly.name_pattern` attribute to track gaps? Further, how are those scenarios not susceptible to the problems we discussed previously?

1sgtpepper: oh my bad give me a sec

include/llzk/Dialect/Polymorphic/IR/Ops.h
```diff
@@ -17,11 +17,20 @@
...
 #include <mlir/Interfaces/InferTypeOpInterface.h>
 
+#include <llvm/ADT/StringRef.h>
+
 // Include TableGen'd declarations
 #include "llzk/Dialect/Polymorphic/IR/OpInterfaces.h.inc"
 
 namespace llzk::polymorphic {
 
+/// Transform-carried metadata preserving the literal chunks of a partial template's logical name.
```
tim-hoffman:
```suggestion
/// Metadata carried across transformation passes preserving the literal chunks of a partially-instantiated template's logical name.
```

include/llzk/Dialect/Polymorphic/IR/Ops.h
```diff
@@ -17,11 +17,20 @@
...
+///
+/// When present, the attribute has P+1 StringAttr chunks around the P current template
+/// parameters. The gap between chunks i and i+1 represents current parameter i. The chunks are
```
tim-hoffman:
```suggestion
/// parameters. The gap between chunks `i` and `i+1` represents current parameter `i`. The chunks are
```

test/Transforms/Flattening/instantiate_name_encoding.llzk
```diff
@@ -79,7 +79,8 @@ module attributes {llzk.lang} {
 // -----
 
 module attributes {llzk.lang} {
-  // A previously partial template is refined from literal chunks, not from its physical name.
+  // A previously partial template is refined from `poly.name_pattern` chunks, not from its
+  // physical name.
```
tim-hoffman:
```suggestion
  // The name of a partially instantiated template is refined from `poly.name_pattern` chunks,
  // not from its physical name.
```

tim-hoffman (changes requested):
Implementation looks good but I'm requesting a few documentation clarifications.

tim-hoffman (approved):
LGTM; thanks for your contribution!

PR #636, opened 2026-07-23
Relax `isConcrete` type restriction for replacements in `llzk-infer-tvar` pass

tim-hoffman:
Extracted from https://github.com/project-llzk/llzk-lib/pull/565 where I noticed the `isConcrete` check was overly restrictive for what this pass is capable of.

### DIFF
--- a/include/llzk/Util/TypeHelper.h
+++ b/include/llzk/Util/TypeHelper.h
@@ -111,6 +111,14 @@ bool isValidArrayType(mlir::Type type);
 /// - `StructType` with parameters if `allowStructParams==false`
 bool isConcreteType(mlir::Type type, bool allowStructParams = true);
 
+/// Return `false` if the type contains a `TypeVarType`.
+///
+/// This is weaker than `isConcreteType`: array dimensions may be symbols or affine maps, and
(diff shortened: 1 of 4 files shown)


iangneal (approved):

shankarapailoor:
My concern is that when you substitute t-var free types which have types that are value-parameteric across template boundaries. Take something like this:

```
poly.template @B {
  poly.param @N : index
  poly.param @T : !poly.tvar<@T>

  function.def @f(%arg: !poly.tvar<@T>) { ... }
  // somewhere inside: a cast proving  @T := !array.type<@N x felt>
}

poly.template @A {
  poly.param @U : !poly.tvar<@U>

  function.def @g(%x: !poly.tvar<@U>) {
    // A forwards its own tvar @U as B's template argument @T:
    function.call @B::@f<[7, !poly.tvar<@U>]>(%x) ...
  }
}
```

Would you get that `@U = array.type<@N x felt>` where `@N` references @B's parameter? Couldn't this cause issues?

tim-hoffman:
> My concern is that when you substitute t-var free types which have types that are value-parameteric across template boundaries.

fixed: https://github.com/project-llzk/llzk-lib/pull/639

PR #638, opened 2026-07-23
Reject felt template table offsets

1sgtpepper:
Addresses the table-offset review feedback from #604.

Reject felt template values when materializing symbolic `struct.readm` table offsets; integer template offsets remain unchanged. Document the index-only rule.

Covers a rejected felt-valued offset and the existing accepted integer path.

Validation:
CI

### DIFF
--- a/changelogs/unreleased/fix__felt-template-table-offsets.yaml
+++ b/changelogs/unreleased/fix__felt-template-table-offsets.yaml
@@ -1,2 +1,2 @@
 fixed:
-  - Reject felt template values used as index-only member table offsets
+  - In `llzk-flatten` pass, reject felt template values used as index-only member table offsets
--- a/changelogs/unreleased/fix__felt-template-table-offsets.yaml
+++ b/changelogs/unreleased/fix__felt-template-table-offsets.yaml
@@ -1,0 +1,2 @@
+fixed:
+  - Reject felt template values used as index-only member table offsets
--- a/include/llzk/Dialect/Polymorphic/Transforms/TransformationPasses.td
+++ b/include/llzk/Dialect/Polymorphic/Transforms/TransformationPasses.td
@@ -61,7 +61,6 @@ def FlatteningPass : LLZKPass<"llzk-flatten"> {
     - Replace parameterized structs with flattened (i.e., no parameter)
       versions of those structs based on requested return type at calls
       to `compute()` functions and unroll loops
-    - Materialize symbolic member table offsets only from integer template values
     - Unroll loops
   }];
   // Implementation note: These options should be kept in sync with
--- a/include/llzk/Dialect/Polymorphic/Transforms/TransformationPasses.td
+++ b/include/llzk/Dialect/Polymorphic/Transforms/TransformationPasses.td
@@ -61,6 +61,7 @@ def FlatteningPass : LLZKPass<"llzk-flatten"> {
     - Replace parameterized structs with flattened (i.e., no parameter)
       versions of those structs based on requested return type at calls
       to `compute()` functions and unroll loops
+    - Materialize symbolic member table offsets only from integer template values
     - Unroll loops
   }];
   // Implementation note: These options should be kept in sync with
--- a/lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
+++ b/lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
@@ -616,6 +616,14 @@ class StructCloner {
       return success();
     }
 
+    LogicalResult handleDefaultRewrite(
+        Attribute, MemberReadOp op, OpAdaptor, ConversionPatternRewriter &, Attribute a
+    ) const override {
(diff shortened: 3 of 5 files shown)


include/llzk/Dialect/Polymorphic/Transforms/TransformationPasses.td
```diff
@@ -61,6 +61,7 @@ def FlatteningPass : LLZKPass<"llzk-flatten"> {
     - Replace parameterized structs with flattened (i.e., no parameter)
       versions of those structs based on requested return type at calls
       to `compute()` functions and unroll loops
+    - Materialize symbolic member table offsets only from integer template values
```
tim-hoffman: It's such a minor detail. Don't mention it here.

changelogs/unreleased/fix__felt-template-table-offsets.yaml
```diff
@@ -0,0 +1,2 @@
+fixed:
+  - Reject felt template values used as index-only member table offsets
```
tim-hoffman:
```suggestion
  - In `llzk-flatten` pass, reject felt template values used as index-only member table offsets
```

test/Transforms/Flattening/instantiate_function_table_offset.llzk
```diff
@@ -0,0 +1,34 @@
...
+
+  function.def @main(%data: !struct.type<@Data>) -> !felt.type<"bn128"> {
+    %value = function.call @OffsetTemplate::@read<[2]>(%data) : (!struct.type<@Data>) -> !felt.type<"bn128">
+    function.return %value : !felt.type<"bn128">
+  }
+}
+// CHECK-LABEL: function.def @OffsetTemplate_2_read(
+// CHECK:         struct.readm {{.*}} {tableOffset = 2 : index}
+// CHECK-LABEL: function.def @main(
+// CHECK:         function.call @OffsetTemplate_2_read(
```
tim-hoffman: please replace this with the entire output generated by the script

tim-hoffman (changes requested):

lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
```diff
@@ -502,12 +502,60 @@ evaluateTemplateExprs(TemplateOp templateOp, DenseMap<Attribute, Attribute> &par
...
 }
 
-namespace Step1_InstantiateStructs {
-
 static inline bool tableOffsetIsntSymbol(MemberReadOp op) {
   return !llvm::isa_and_present<SymbolRefAttr>(op.getTableOffset().value_or(nullptr));
 }
 
+/// Materialize symbolic member table offsets only from integer template bindings. Member tables are
+/// index-addressed, so other concrete attribute kinds must diagnose instead of being coerced.
```
iangneal: I assume this means "other concrete attributes emit diagnostics"? I'm not sure what the attributes themselves would be diagnosing.

lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
```diff
@@ -502,12 +502,60 @@ evaluateTemplateExprs(TemplateOp templateOp, DenseMap<Attribute, Attribute> &par
...
+class ClonedMemberReadOpPattern
```
iangneal, line 511: This PR diff is larger than it needs to be since this class was renamed `ClonedMemberReadOpPattern` from `ClonedStructMemberReadOpPattern` (unnecessary rename) and moved earlier in the file. Unless there's a compelling reason to do this, please move it back and revert the name change to minimize the diff.

iangneal (dismissed):
Logic is fine but I have a few nits.

lib/Dialect/Polymorphic/Transforms/FlatteningPass.cpp
1sgtpepper: I kept the shared pattern because both struct and function cloning use it.

iangneal: Well it was called `ClonedStructMemberReadOpPattern` because it targets the struct member read op (`MemberReadOp`), no? And there's still no reason to reorder it in the file.

tim-hoffman: Unless it was a matter of moving it outside one of the nested namespaces to be accessible in both locations.

tim-hoffman: Yeah, looks like it was moved outside `namespace Step1_InstantiateStructs` so it's accessible in the step2 namespace as well.

iangneal: Oh, gotcha. Well, then this is probably fine.

tim-hoffman (approved):
LGTM

Issue #585, opened 2026-07-06
PCL backend support for Noir->LLZK compiled code


matthew-a-klein:
Implement PCL backend to fully support noir-> llzk lowering. See [this issue](https://github.com/project-llzk/noir_llzk/issues/53).

Plan is to implement in following stages: 
1. Inline top level call functions (Blackbox functions are implemented as calls to module level functions)
2. Add bit decomposition helper functions and algebraise non native field operations as bit decompositions using new helpers
4. Arithmetise `scf.if` operations using select polynomials with a new pre-lowering pass.

PR #640, opened 2026-07-23
Add pass to algebraize felt operations

krushimir:
Implements stage 2 of #585.

Adds `-llzk-algebraize-felt-ops`, an LLZK pre-lowering pass for functions marked `function.allow_constraint`. It lowers unsupported felt bitwise, shift, inverse, division, unsigned division, and modulo operations into basic felt arithmetic, nondeterministic witnesses, bit decompositions, and equality constraints.

The pass reuses decompositions, uses known width bounds where available, and rejects unsupported division/modulo cases rather than producing unsound constraints.

Tested with the AlgebraizeFeltOps and PCLLowering lit suites (15 tests).

### DIFF
--- a/include/llzk/Transforms/LLZKTransformationPasses.td
+++ b/include/llzk/Transforms/LLZKTransformationPasses.td
@@ -127,6 +127,40 @@ def EnforceNoMemberOverwritePass : LLZKPass<"llzk-enforce-no-overwrite"> {
   }];
 }
 
+def AlgebraizeFeltOpsPass : LLZKPass<"llzk-algebraize-felt-ops"> {
+  let summary = "Rewrite non-native felt ops in @constrain bodies into "
+                "algebraic constraints";
+  let description = [{
+    Rewrites felt operations that have no direct polynomial form —
+    `felt.bit_and`, `felt.bit_or`, `felt.bit_xor`, `felt.bit_not`,
+    `felt.shl`, `felt.shr`, `felt.inv`, `felt.div`, `felt.uintdiv`, and
+    `felt.umod` — inside constraint-bearing functions (those marked
+    `function.allow_constraint`, e.g. `@constrain` and `@product`) into
+    native felt arithmetic over `llzk.nondet` witnesses pinned by
+    `constrain.eq`.
+
+    Bitwise and shift operations are expressed through checked bit
+    decompositions (booleanity, weighted-sum equality, and a range check
+    against the prime when decomposing at full field width). A value proven
+    smaller than `2^k` — by an asserted range guard `x < c`, by constant or
+    integer-cast provenance, or by propagation through `felt.add`/`felt.mul`
+    — is decomposed with `k` bits instead of the full field width.
+    Decompositions are cached and shared between operations on the same
+    value.
+
+    `felt.inv`/`felt.div` introduce a witness `w` with `v*w == 1`
+    (unsatisfiable for `v == 0`). `felt.uintdiv`/`felt.umod` support only
+    constant power-of-two divisors, taken as slices of the dividend's
+    decomposition; a sound encoding for dynamic divisors would require a
+    multiprecision product argument.
+
+    Values are not analyzed across `function.call`s, so running after call
+    inlining gives the tightest decompositions. Run before a constraint
+    backend (e.g. `-llzk-to-pcl`), which then only sees native felt
+    arithmetic.
+  }];
+}
+
 def WhileToForPass : LLZKPass<"llzk-while-to-for"> {
   let summary =
       "Converts scf.while loops to equivalent scf.for loops when possible";
--- a/lib/Transforms/LLZKAlgebraizeFeltOpsPass.cpp
+++ b/lib/Transforms/LLZKAlgebraizeFeltOpsPass.cpp
@@ -1,0 +1,683 @@
+//===-- LLZKAlgebraizeFeltOpsPass.cpp ---------------------------*- C++ -*-===//
+//
+// Part of the LLZK Project, under the Apache License v2.0.
+// See LICENSE.txt for license information.
+// Copyright 2026 Project LLZK
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file implements the `-llzk-algebraize-felt-ops` pass.
+///
+//===----------------------------------------------------------------------===//
+
+#include "llzk/Dialect/Bool/IR/Ops.h"
+#include "llzk/Dialect/Cast/IR/Ops.h"
+#include "llzk/Dialect/Constrain/IR/Ops.h"
+#include "llzk/Dialect/Felt/IR/Attrs.h"
+#include "llzk/Dialect/Felt/IR/Ops.h"
+#include "llzk/Dialect/Function/IR/Ops.h"
+#include "llzk/Dialect/LLZK/IR/Ops.h"
+#include "llzk/Transforms/LLZKTransformationPasses.h"
+#include "llzk/Util/DynamicAPIntHelper.h"
+#include "llzk/Util/Field.h"
+
+#include <mlir/IR/BuiltinOps.h>
+#include <mlir/IR/Matchers.h>
+
+#include <llvm/ADT/APInt.h>
+#include <llvm/ADT/DenseMap.h>
+#include <llvm/ADT/STLExtras.h>
+#include <llvm/ADT/SmallVector.h>
+#include <llvm/ADT/TypeSwitch.h>
+
+#include <deque>
+#include <optional>
+#include <utility>
+
+namespace llzk {
+#define GEN_PASS_DEF_ALGEBRAIZEFELTOPSPASS
+#include "llzk/Transforms/LLZKTransformationPasses.h.inc"
+} // namespace llzk
+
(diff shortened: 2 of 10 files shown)


iangneal:
@codex review

lib/Transforms/LLZKAlgebraizeFeltOpsPass.cpp
```diff
@@ -0,0 +1,683 @@
...
+  FailureOr<Value> rewriteBitwiseNot(NotFeltOp op);
+  FailureOr<Value> rewriteShl(ShlFeltOp op);
+  FailureOr<Value> rewriteShr(ShrFeltOp op);
+  FailureOr<Value> rewriteDivModPow2(Operation *op, bool wantQuotient);
+};
+
+const llvm::APInt &ConstrainAlgebraizer::primeOf(Type type) {
+  auto [it, inserted] = primes.try_emplace(type);
+  if (inserted) {
+    FeltType feltType = llvm::cast<FeltType>(type);
```
raghav198, line 134: This cast might fail if the `type` passed in isn't actually a `!felt.type`, right? Can we have this method just take a `FeltType` and require the caller to check the cast instead?

raghav198: Similar comments elsewhere `llvm::cast<FeltType>(...)` is used

test/Transforms/AlgebraizeFeltOps/algebraize_bitwise_pass.llzk
```diff
@@ -0,0 +1,152 @@
...
+!F = !felt.type<"koalabear">
+
+module attributes {llzk.lang} {
+  struct.def @MixedXor {
+    struct.member @out : !F {llzk.pub}
+    function.def @compute(%a: !F, %b: !F) -> !struct.type<@MixedXor> {
+      %self = struct.new : <@MixedXor>
+      function.return %self : !struct.type<@MixedXor>
+    }
+    function.def @constrain(%self: !struct.type<@MixedXor>, %a: !F, %b: !F) attributes {function.allow_constraint, function.allow_non_native_field_ops} {
```
0xddom, line 100: Since this pass removes all the operations enabled by this attribute, we could also remove it from the function.

test/Transforms/AlgebraizeFeltOps/algebraize_region_block_arg.llzk
```diff
@@ -0,0 +1,29 @@
+// RUN: llzk-opt -llzk-algebraize-felt-ops %s > /dev/null
```
0xddom, line 1: Is this test a TODO?

lib/Transforms/LLZKAlgebraizeFeltOpsPass.cpp
0xddom: Seconding this suggestion

lib/Transforms/LLZKAlgebraizeFeltOpsPass.cpp
```diff
@@ -0,0 +1,683 @@
...
+
+  ArrayRef<Value> sBits = getBits(loc, op.getRhs());
+  Value one = feltConst(loc, type, 1);
+  llvm::DynamicAPInt primeDyn = toDynamicAPInt(prime);
+  llvm::DynamicAPInt curPow2(2);
+
+  Value pow2s = one;
+  for (unsigned i = 0, e = sBits.size(); i < e; ++i) {
+    llvm::APInt cMinus1 = toExactWidthAPInt(curPow2 - 1, constBits);
+    Value scaled = b.create<MulFeltOp>(loc, sBits[i], feltConst(loc, type, cMinus1));
```
0xddom, line 495: nit: `i` is not used anywhere else in this loop

```suggestion
  for (auto &sBit : sBits) {
    llvm::APInt cMinus1 = toExactWidthAPInt(curPow2 - 1, constBits);
    Value scaled = b.create<MulFeltOp>(loc, sBit, feltConst(loc, type, cMinus1));
```

lib/Transforms/LLZKAlgebraizeFeltOpsPass.cpp
```diff
@@ -0,0 +1,683 @@
...
+    return it->second;
+  }
+  return fullWidthOf(v.getType());
+}
+
+/// Records the bound proven by an asserted range guard
+/// `constrain.eq(cast.tofelt(bool.cmp lt/le/gt/ge(x, c)), 1)` — the pattern
+/// frontends emit to pin blackbox input widths. Only guards at the top level
+/// of the body may be harvested: a guard inside a conditional region holds
+/// only on its branch.
```
0xddom, line 207: What's the purpose of these guards in relation to the ops converted by this pass? Because, when converting an op inside a conditional block, wouldn't we be missing a range constraint for that op if it was also defined inside the same block?

lib/Transforms/LLZKAlgebraizeFeltOpsPass.cpp
```diff
@@ -0,0 +1,683 @@
...
+/// Combinator inverse of `decomposeBits`: returns `sum(bits[i] * 2^i)`.
+/// Emits no assertions; callers are responsible for any booleanity or range
+/// constraints on the input bits. Precondition: `bits` is non-empty.
+Value ConstrainAlgebraizer::recomposeBits(Location loc, ArrayRef<Value> bits) {
+  assert(!bits.empty() && "recomposeBits requires at least one bit");
+  Type type = bits[0].getType();
+  llvm::APInt weight(primeOf(type).getBitWidth(), 1);
+
+  Value acc = bits[0];
+  weight <<= 1;
```
0xddom, line 309: Why not set `weight` directly to 2?

lib/Transforms/LLZKAlgebraizeFeltOpsPass.cpp
```diff
@@ -0,0 +1,683 @@
...
+  for (unsigned i = 1, e = bits.size(); i < e; ++i) {
+    Value term = b.create<MulFeltOp>(loc, bits[i], feltConst(loc, type, weight));
```
0xddom, line 311: nit: similar comment about `i`

```suggestion
  for (auto &bit : bits.drop_front()) {
    Value term = b.create<MulFeltOp>(loc, bit, feltConst(loc, type, weight));
```

lib/Transforms/LLZKAlgebraizeFeltOpsPass.cpp
```diff
@@ -0,0 +1,683 @@
...
+
+LogicalResult ConstrainAlgebraizer::run() {
+  // Range guards may sit anywhere in the top-level body: constraints form
+  // one conjunction, so a guard bounds its value wherever the value is used.
+  for (Operation &op : fn.getBody().front()) {
+    if (auto eq = llvm::dyn_cast<EmitEqualityOp>(&op)) {
+      harvestRangeGuard(eq.getLhs(), eq.getRhs());
+      harvestRangeGuard(eq.getRhs(), eq.getLhs());
+    }
+  }
```
0xddom, line 591:
```suggestion
  fn.walk([this](EmitEqualityOp eq) {
    harvestRangeGuard(eq.getLhs(), eq.getRhs());
    harvestRangeGuard(eq.getRhs(), eq.getLhs());
  });
```

lib/Transforms/LLZKAlgebraizeFeltOpsPass.cpp
```diff
@@ -0,0 +1,683 @@
...
+
+  // One pass in program order: operands are defined before uses, so bounds
+  // are propagated before the ops consuming them are rewritten, and a
+  // rewritten op's result bound (set by cacheResultBits) feeds later ops.
+  // The snapshot keeps the loop off the ops the rewrites create.
+  SmallVector<Operation *> ops;
+  fn.walk([&](Operation *op) { ops.push_back(op); });
```
0xddom, line 598: nit: prefer to pass only the bindings actually used by the lambda

```suggestion
  fn.walk([&ops](Operation *op) { ops.push_back(op); });
```

lib/Transforms/LLZKAlgebraizeFeltOpsPass.cpp
```diff
@@ -0,0 +1,683 @@
...
+
+  for (Operation *op : ops) {
```
0xddom, line 600: Instead of iterating over all the operations twice you could run the algorithm in the first pass, accumulating the to-be-removed ops along with their replacement. Then do another loop that actually removes and replaces the ops.

include/llzk/Transforms/LLZKTransformationPasses.td
```diff
@@ -127,6 +127,40 @@ def EnforceNoMemberOverwritePass : LLZKPass<"llzk-enforce-no-overwrite"> {
   }];
 }
 
+def AlgebraizeFeltOpsPass : LLZKPass<"llzk-algebraize-felt-ops"> {
```
0xddom, line 130: The pass could be made an operation pass for `function.def`. We don't need a whole-world pass for this and will allow running in parallel.

tim-hoffman:
@krushimir I'm going to mark this as a draft for now but when you've resolved conflicts and the feedback, please mark it ready for review again.

krushimir:
> 

Thanks! I believe @matthew-a-klein is taking over this one too.

PR #643, opened 2026-07-24
Arithmetize constraint-side `scf.if` operations

krushimir:
Implements stage 3 of #585.

Adds `-llzk-arithmetize-scf-if`, which replaces constraint-side `scf.if` operations with select polynomials. Felt results use `else + condition * (then - else)`, while branch constraints are gated so inactive branches do not apply. Nested conditionals are supported.

Because both branches execute after rewriting, branch operations must be safe to speculate. Unsupported operations and non-felt results or constraints are rejected. Run `-llzk-algebraize-felt-ops` first for felt division and non-native operations.

Also fixes PCL legalization for operations outside constraint functions, such as `scf` operations in `@compute`.

Tested with the ArithmetizeSCFIf and PCLLowering lit suites. Together with #586 and #640, the full pipeline lowers all eight `noir_llzk` examples, including `multi_scalar_mul`.


iangneal:
@codex review

0xddom:
> Also fixes PCL legalization for operations outside constraint functions, such as `scf` operations in `@compute`.

The PCL conversion pass shouldn't care about what goes on inside `@compute`, so I don't fully understand this point.

lib/Transforms/LLZKArithmetizeSCFIfPass.cpp
```diff
@@ -0,0 +1,288 @@
...
+    ifOp.emitError("scf.if arithmetization supports only felt results");
+    return failure();
+  }
+  if (failed(validateBranch(ifOp, ifOp.getThenRegion().front()))) {
+    return failure();
+  }
+  if (!ifOp.getElseRegion().empty() && failed(validateBranch(ifOp, ifOp.getElseRegion().front()))) {
+    return failure();
+  }
+  return success();
```
0xddom, line 149: `success` and `failure` accept a `bool` as optional argument so this could be written as something similar to this:

```suggestion
  return failure(
    failed(validateBranch(ifOp, ifOp.getThenRegion().front())) || 
    (!ifOp.getElseRegion().empty() && failed(validateBranch(ifOp, ifOp.getElseRegion().front())))
  );
```

lib/Transforms/LLZKArithmetizeSCFIfPass.cpp
```diff
@@ -0,0 +1,288 @@
...
+    ifOp.emitError("expected each scf.if branch to end with scf.yield");
+    return failure();
+  }
+
+  for (Operation &op : block.without_terminator()) {
+    if (llvm::isa<mlir::scf::IfOp>(op)) {
+      continue;
+    }
+    if (auto eq = llvm::dyn_cast<EmitEqualityOp>(op)) {
+      if (!llvm::isa<FeltType>(eq.getLhs().getType())) {
```
0xddom, line 113: Is not necessarily the case that `eq.getLhs().getType() == eq.getRhs().getType()` so this check should probably handle both.

lib/Transforms/LLZKArithmetizeSCFIfPass.cpp
```diff
@@ -0,0 +1,288 @@
...
+    }
+    if (requiresFeltAlgebraization(&op)) {
+      op.emitError("felt division and non-native operations must be lowered before scf.if");
+      return failure();
+    }
+    if (op.getNumRegions() != 0 || op.getNumSuccessors() != 0 ||
+        !isSafeToExecuteUnconditionally(&op)) {
+      op.emitError() << "cannot arithmetize scf.if branch operation '" << op.getName()
+                     << "' because executing both branches may change its effects";
+      return failure();
```
0xddom, line 127: The result of this expression can be converted into a `LogicalResult`. 

```suggestion
      return op.emitError() << "cannot arithmetize scf.if branch operation '" << op.getName()
                     << "' because executing both branches may change its effects";
```

There are other places where this could be applied as well.

lib/Transforms/LLZKArithmetizeSCFIfPass.cpp
```diff
@@ -0,0 +1,288 @@
...
+  SmallVector<mlir::scf::IfOp> ifOps;
+  fn.walk<WalkOrder::PostOrder>([&](mlir::scf::IfOp ifOp) { ifOps.push_back(ifOp); });
+
+  // Validate the whole function before changing it. Nested scf.if operations
+  // are allowed here because the rewrite list is already bottom-up.
+  for (mlir::scf::IfOp ifOp : ifOps) {
+    if (failed(validate(ifOp))) {
+      return failure();
+    }
+  }
```
0xddom, line 256: You can use `llvm::any_of` for this. Something like

```
if (llvm::any_of(ifOps, [this](mlir::scf::IfOp ifOp) { return failed(validate(ifOp)); })) {
  return failure();
}
```

include/llzk/Transforms/LLZKTransformationPasses.td
```diff
@@ -127,6 +127,33 @@ def EnforceNoMemberOverwritePass : LLZKPass<"llzk-enforce-no-overwrite"> {
   }];
 }
 
+def ArithmetizeSCFIfPass : LLZKPass<"llzk-arithmetize-scf-if"> {
```
0xddom, line 130: We can also make this pass an op pass for `function.def`. The same rationale in #640 applies here.

0xddom (commented):
Still going through the code. I have a first batch of comments.

V-k-h:
Built this branch and ran it on a real Noir `multi_scalar_mul` circuit: `--llzk-arithmetize-scf-if` cleanly dissolves all ~2800 constraint-side `scf.if` (2801 → 0), with end-to-end lowering then gated only on the felt-op support in #640. Nice to see this land — the `else + condition * (then − else)` approach is exactly right.

Minor, and pre-existing (not introduced here): with the pass built but not invoked, `--llzk-full-r1cs-lowering` still *segfaults* in `lowerPolyToR1CS` on a data-dependent `scf.if`, whereas `--llzk-to-pcl` rejects the same input cleanly (`failed to legalize operation 'arith.select'`). Happy to file that separately if it's useful rather than something you're already handling as part of the pipeline wiring.

0xddom:
> Built this branch and ran it on a real Noir `multi_scalar_mul` circuit: `--llzk-arithmetize-scf-if` cleanly dissolves all ~2800 constraint-side `scf.if` (2801 → 0), with end-to-end lowering then gated only on the felt-op support in #640. Nice to see this land — the `else + condition * (then − else)` approach is exactly right.
> 
> Minor, and pre-existing (not introduced here): with the pass built but not invoked, `--llzk-full-r1cs-lowering` still _segfaults_ in `lowerPolyToR1CS` on a data-dependent `scf.if`, whereas `--llzk-to-pcl` rejects the same input cleanly (`failed to legalize operation 'arith.select'`). Happy to file that separately if it's useful rather than something you're already handling as part of the pipeline wiring.

Could you create issues for those two errors? The latter sounds like the `--llzk-to-pcl` is being fed ops that doesn't support (`arith.select`, which iirc is a sort of ternary operator). If those exist as a consequence of these new passes then if should be fixed before merging

V-k-h:
Will do — I'll open both.

One clarification on provenance, since it bears on your "fix before merging" point: the `arith.select` isn't emitted by this PR's pass. `--canonicalize` alone rewrites the input's *data-dependent* `scf.if` into `arith.select` (confirmed on the minimal repro below), so it's already present before any of the #585-stack passes run — both failures above were with `--llzk-arithmetize-scf-if` built but **not** invoked. When the pass *is* invoked it's exactly what removes them (the 2801 → 0 on the real `multi_scalar_mul` circuit reported above). So neither error is a consequence of the new passes; they're pre-existing gaps in the default `--llzk-full-r1cs-lowering` / `--llzk-to-pcl` pipelines (which don't yet wire in the new pass), and #643 + #640 is the resolution rather than the cause — nothing here should block merging.

Both issues will carry a 38-line minimal reproducer and the exact invocations.

Filed: #664 (`--llzk-full-r1cs-lowering` / `lowerPolyToR1CS` segfault) and #665 (`--llzk-to-pcl` / `arith.select`).

tim-hoffman:
I'm going to mark this as a draft for now but when you've resolved conflicts and the feedback, please mark it ready for review again.

krushimir:
> I'm going to mark this as a draft for now but when you've resolved conflicts and the feedback, please mark it ready for review again.

Thanks! Sorry for the silence, I believe @matthew-a-klein is taking over this.

0xddom:
Thinking about this PR and its relation with #665.

- The PCL dialect does not support that kind of ternary operator and as far as I understand we don't plan to (@shankarapailoor can give more insight here). 
- If `arith.select` is emitted by the canonicalizer from some `scf.if` ops, that means that the former is the canonical form of the latter.

If `arith.select` is an acceptable form of representing conditionals, that means that passes working with `scf.if` should support it as well. I think that passes in general should expect IR to be on its canonical form (with the exception of complicated multi-stage pipelines, but those should be rare, adequately justified, and explicitly documented). The user can always run `--canonicalize` before feeding the IR to the pass. 

That leads me to the following question: are these kind of _convertible-to_-`arith.select` `scf.if` ops part of the conditionals that this pass targets? If so, then we need to add support for arithmetizing `arith.select` ops as well.

PR #674, opened 2026-08-05
code cleanup and formatting

tim-hoffman:
Just code style. Most changes where just find-replace the `mlir::` prefix since I added the "using namespace"


shankarapailoor (approved):
LGTM

PR #676, opened 2026-08-06
Assorted PCL backend bugfixes

0xddom:
Fixes crashes that occurred in the LLZK->PCL conversion pass and in the PCL translation:

- Ops not supported by the LLZK->PCL pass that are located in functions not targetted by the lowering are now ignored.
- Handle support for calls to free functions from the constrain functions
- Fix an infinite loop while printing some unary operators.
- Improve performance of the translator by caching s-expressions already generated.

### DIFF
--- a/backends/pcl/lib/Conversion/PCLLoweringPass.cpp
+++ b/backends/pcl/lib/Conversion/PCLLoweringPass.cpp
@@ -46,6 +46,7 @@
 
 #include <mlir/Dialect/Arith/IR/Arith.h>
 #include <mlir/Dialect/Func/IR/FuncOps.h>
+#include <mlir/Dialect/SCF/IR/SCF.h>
 #include <mlir/IR/Attributes.h>
 #include <mlir/IR/Builders.h>
 #include <mlir/IR/BuiltinAttributes.h>
@@ -64,6 +65,7 @@
 #include <llvm/ADT/APInt.h>
 #include <llvm/ADT/DenseMap.h>
 #include <llvm/ADT/DenseMapInfo.h>
+#include <llvm/ADT/DenseSet.h>
 #include <llvm/ADT/EquivalenceClasses.h>
 #include <llvm/ADT/SmallVector.h>
 #include <llvm/ADT/SmallVectorExtras.h>
@@ -94,7 +96,7 @@ using namespace llzk::component;
 namespace {
 
 /// Returns a flat representation of the fully qualified name of the struct.
-static std::string flatStructName(StructDefOp op) {
+template <typename Op> static std::string flatStructName(Op op) {
   auto fqn = op.getFullyQualifiedName();
   std::string name;
   llvm::raw_string_ostream o(name);
@@ -394,7 +396,7 @@ template <typename T, typename Fn> SmallVector<T> mapOutputMembers(StructDefOp o
   return out;
 }
 
-/// Converts `function.return` ops into pcl return ops.
+/// Converts `function.return` ops inside `@constrain` functions into func return ops.
 struct ConvertReturnOp : public OpConversionPattern<ReturnOp> {
   using OpConversionPattern<ReturnOp>::OpConversionPattern;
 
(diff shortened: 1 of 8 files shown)


backends/pcl/lib/Conversion/PCLLoweringPass.cpp
```diff
@@ -94,7 +96,7 @@ using namespace llzk::component;
 namespace {
 
 /// Returns a flat representation of the fully qualified name of the struct.
-static std::string flatStructName(StructDefOp op) {
+template <typename Op> static std::string flatStructName(Op op) {
```
tim-hoffman: I'm guessing you changed this to allow `FuncDefOp` and/or `ContractOp` since they have `getFullyQualifiedName()` but the function name should be changed as well if it's not specific to struct.

tim-hoffman: Could also consider directly calling `llzk::getFullyQualifiedName` which is the implementation of all of those.

backends/pcl/lib/Conversion/PCLLoweringPass.cpp
```diff
@@ -413,6 +415,22 @@ struct ConvertReturnOp : public OpConversionPattern<ReturnOp> {
...
 };
 
+/// Converts `function.return` ops inside free functions into func return ops.
+struct ConvertFreeFuncReturnOp : public OpConversionPattern<ReturnOp> {
+  using OpConversionPattern<ReturnOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(ReturnOp op, OpAdaptor adaptor, ConversionPatternRewriter &rewriter) const override {
+    auto structDefOp = op->getParentOfType<StructDefOp>();
+    if (structDefOp) {
```
tim-hoffman:
```suggestion
    if (op->getParentOfType<StructDefOp>()) {
```

backends/pcl/lib/Conversion/PCLLoweringPass.cpp
```diff
@@ -590,6 +677,56 @@ struct ConvertConstrainCall : public OpConversionPattern<CallOp> {
...
+    SmallVector<Type> resultTypes(op.getNumResults(), pcl::FeltType::get(getContext()));
+    auto calleeName = flatStructName(callee->get());
+    rewriter.replaceOpWithNewOp<func::CallOp>(
+        op, calleeName, TypeRange(resultTypes), adaptor.getArgOperands()
+    );
+    return success();
+  }
+};
+
+class RemoveFreeFunction : public OpConversionPattern<FuncDefOp> {
```
tim-hoffman, line 705: document

backends/pcl/lib/Conversion/PCLLoweringPass.cpp
```diff
@@ -640,11 +782,18 @@ static void populateStep3ConversionPatterns(
 ///
 /// An operation is legal in Step 1 if its located outside the
 /// `constrain` function of a struct.
-static bool isStep1LegalOp(Operation *op) {
+static bool isStep1LegalOp(Operation *op, llvm::DenseSet<FuncDefOp> &usedFreeFunctions) {
   auto structDefOp = op->getParentOfType<StructDefOp>();
   if (!structDefOp) {
```
tim-hoffman: I think cleaner to invert the if, like `if (auto structDefOp = op->getParentOfType<StructDefOp>()) {`, since the only use of `structDefOp` is now at the very end of the function.

backends/pcl/lib/Conversion/PCLLoweringPass.cpp
```diff
@@ -656,28 +805,33 @@ static bool isStep1LegalOp(Operation *op) {
...
 /// Populates the conversion target with the legallity expected of step 2 of the conversion.
 static void populateStep2ConversionTarget(ConversionTarget &target) {
   target.addLegalDialect<pcl::PCLDialect>();
   target.addLegalOp<ModuleOp, func::FuncOp, func::CallOp, func::ReturnOp>();
-  target.addIllegalOp<StructDefOp>();
+  target.addIllegalOp<StructDefOp, FuncDefOp>();
 }
 
-/// Populates the conversion target with the legalluty expected of step 3 of the conversion.
+/// Populates the conversion target with the legallity expected of step 3 of the conversion.
```
tim-hoffman:
```suggestion
/// Populates the conversion target with the legality expected of step 3 of the conversion.
```

backends/pcl/lib/Conversion/PCLLoweringPass.cpp
```diff
@@ -888,28 +1042,72 @@ class PassImpl : public pcl::impl::PCLLoweringPassBase<PassImpl> {
...
+          return;
+        }
+        funcs.insert(callee->get());
+        WL.push_back(callee->get());
+      });
+    });
+
+    // Complete the graph using the worklist.
+    while (!WL.empty()) {
+      auto next = WL[WL.size() - 1];
```
tim-hoffman:
```suggestion
      auto next = WL.back();
```

tim-hoffman (changes requested):

backends/pcl/lib/Conversion/PCLLoweringPass.cpp
0xddom: Yeah, but the difference with that method is that I want a single string for the name since I flatten all the submodules. I renamed it to `flatFullyQualifiedName` which should be pretty indicative.

tim-hoffman: I meant just to replace the `op.getFullyQualifiedName()` line and then you can avoid the template and use `mlir::SymbolOpInterface` as the parameter type. But either way is fine w.r.t. that.

tim-hoffman (approved):
LGTM

tim-hoffman (approved):
LGTM

PR #677, opened 2026-08-07
Add trim pass

0xddom:
A pass for reducing the size of PCL expressions to avoid size explosions when printing the IR in lisp.

### DIFF
--- a/backends/pcl/include/pcl/Transforms/TransformationPasses.h
+++ b/backends/pcl/include/pcl/Transforms/TransformationPasses.h
@@ -2,7 +2,7 @@
 //
 // Part of the LLZK Project, under the Apache License v2.0.
 // See LICENSE.txt for license information.
-// Copyright 2025 Veridise Inc.
+// Copyright 2026 Project LLZK
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
--- a/backends/pcl/include/pcl/Transforms/TransformationPasses.h
+++ b/backends/pcl/include/pcl/Transforms/TransformationPasses.h
@@ -1,0 +1,20 @@
+//===-- TransformationPasses.h ----------------------------------*- C++ -*-===//
+//
+// Part of the LLZK Project, under the Apache License v2.0.
+// See LICENSE.txt for license information.
+// Copyright 2025 Veridise Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "llzk/Pass/PassBase.h"
+
+namespace pcl {
+
+#define GEN_PASS_DECL
+#define GEN_PASS_REGISTRATION
+#include "pcl/Transforms/TransformationPasses.h.inc"
+
+} // namespace pcl
--- a/backends/pcl/include/pcl/Transforms/TransformationPasses.td
+++ b/backends/pcl/include/pcl/Transforms/TransformationPasses.td
@@ -2,7 +2,7 @@
 //
 // Part of the LLZK Project, under the Apache License v2.0.
 // See LICENSE.txt for license information.
-// Copyright 2025 Veridise Inc.
+// Copyright 2026 Project LLZK
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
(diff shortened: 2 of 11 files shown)


backends/pcl/include/pcl/Transforms/TransformationPasses.h
```diff
@@ -0,0 +1,20 @@
+//===-- TransformationPasses.h ----------------------------------*- C++ -*-===//
+//
+// Part of the LLZK Project, under the Apache License v2.0.
+// See LICENSE.txt for license information.
+// Copyright 2025 Veridise Inc.
```
tim-hoffman:
```suggestion
// Copyright 2026 Project LLZK
```

backends/pcl/include/pcl/Transforms/TransformationPasses.td
```diff
@@ -0,0 +1,101 @@
+//===-- TransformationPasses.td ----------------------------*- tablegen -*-===//
+//
+// Part of the LLZK Project, under the Apache License v2.0.
+// See LICENSE.txt for license information.
+// Copyright 2025 Veridise Inc.
```
tim-hoffman:
```suggestion
// Copyright 2026 Project LLZK
```

backends/pcl/include/pcl/Transforms/TransformationPasses.td
```diff
@@ -0,0 +1,100 @@
...
+    pcl.assert %12
+    ```
+
+    Which results in the following s-expressions:
+
+    ```
+    (assert (= t0 (+ (+ a b) (* (+ a b) (+ a b)))))
+    (assert (= c t0))
+    (assert (= d t0))
+    (assert (= e t0))
```
tim-hoffman, line 91: Is this not just CSE?

backends/pcl/lib/Transforms/TrimExprSizePass.cpp
```diff
@@ -0,0 +1,157 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "pcl/Dialect/IR/Dialect.h"
+#include "pcl/Dialect/IR/Ops.h"
+#include "pcl/Transforms/TransformationPasses.h"
+
+#include <mlir/IR/Builders.h>
```
tim-hoffman:
```suggestion
#include "pcl/Dialect/IR/Dialect.h"
#include "pcl/Dialect/IR/Ops.h"
#include "pcl/Transforms/TransformationPasses.h"

#include <mlir/Dialect/Func/IR/FuncOps.h>
#include <mlir/IR/Builders.h>
```

backends/pcl/lib/Transforms/TrimExprSizePass.cpp
```diff
@@ -0,0 +1,157 @@
...
+        auto zeroOp = builder.create<pcl::ConstOp>(loc, value);
+        auto cmpEqOp = builder.create<pcl::CmpEqOp>(loc, varOp, zeroOp);
+        auto notOp = builder.create<pcl::NotOp>(loc, cmpEqOp);
+        cut->replaceAllUsesWith(notOp);
+      } else {
+        cut->replaceAllUsesWith(varOp);
+      }
+      if (isa<pcl::BoolType>(cutOpResult.getType())) {
+        cutOpResult = builder.create<pcl::AsFeltOp>(loc, cutOpResult);
+      }
```
tim-hoffman: Is there some reason this cannot be combined with the `if` above?

backends/pcl/include/pcl/Transforms/TransformationPasses.td
tim-hoffman: After reviewing further, I see it's only CSE in theory since it's operating on `pcl.var` rather than SSA values so an existing CSE pass would not capture that.

0xddom: Yeah, it's spiritually a CSE but it's tailored to this specific situation

backends/pcl/lib/Transforms/TrimExprSizePass.cpp
0xddom: Nothing other than I wrote them at different times and didn't think about that

tim-hoffman (approved):
LGTM

PR #682, opened 2026-08-11
Stubbed lowering mode for the LLZK->PCL pass

0xddom:
Refactors the `-llzk-to-pcl` pass to add a stubbed mode. This mode will lower free functions with unsupported ops as stubs. This way we can still get the PCL representation of the supported parts of the circuit.

The default mode is still the full lowering and the stubbed mode needs to be manually enabled.

### DIFF
--- a/backends/pcl/include/pcl/Conversion/ConversionPasses.td
+++ b/backends/pcl/include/pcl/Conversion/ConversionPasses.td
@@ -19,6 +19,15 @@ def PCLLoweringPass : LLZKPass<"llzk-to-pcl"> {
       Transforms LLZK constraints into an equivalent set of constraints expressed in the pcl-mlir dialect.
       This pass currently expects the flattening and inlining passes to be run prior to its use.
     }];
+
+  let options = [Option<
+      "mode", "mode", "LlzkToPclMode", "::pcl::LlzkToPclMode::Full",
+      "Lowering mode (full or stubbed)", [{::llvm::cl::values(
+               clEnumValN(::pcl::LlzkToPclMode::Full, "full",
+                "Lower the IR completely."),
+               clEnumValN(::pcl::LlzkToPclMode::Stubbed, "stubbed",
+                "Replace free functions with unsupported ops with stubs.")
+              )}]>];
 }
 
 #endif // PCL_CONV_CONVERSION_PASSES_TD
--- a/backends/pcl/include/pcl/Conversion/ConversionPasses.td
+++ b/backends/pcl/include/pcl/Conversion/ConversionPasses.td
@@ -26,7 +26,7 @@ def PCLLoweringPass : LLZKPass<"llzk-to-pcl"> {
                clEnumValN(::pcl::LlzkToPclMode::Full, "full",
                 "Lower the IR completely."),
                clEnumValN(::pcl::LlzkToPclMode::Stubbed, "stubbed",
-                "Replace free functions with unsupported ops with stubs.")
+                "Replace free functions containing unsupported ops with stubs.")
               )}]>];
 }
 
--- a/backends/pcl/lib/Conversion/PCLLoweringPass.cpp
+++ b/backends/pcl/lib/Conversion/PCLLoweringPass.cpp
@@ -39,6 +39,7 @@
 #include "llzk/Transforms/LLZKLoweringUtils.h"
 #include "llzk/Util/DynamicAPIntHelper.h"
 #include "llzk/Util/Field.h"
+#include "llzk/Util/SymbolLookup.h"
 
 #include <pcl/Dialect/IR/Dialect.h>
 #include <pcl/Dialect/IR/Ops.h>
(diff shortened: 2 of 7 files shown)


backends/pcl/include/pcl/Conversion/ConversionPasses.td
```diff
@@ -19,6 +19,15 @@ def PCLLoweringPass : LLZKPass<"llzk-to-pcl"> {
...
       This pass currently expects the flattening and inlining passes to be run prior to its use.
     }];
+
+  let options = [Option<
+      "mode", "mode", "LlzkToPclMode", "::pcl::LlzkToPclMode::Full",
+      "Lowering mode (full or stubbed)", [{::llvm::cl::values(
+               clEnumValN(::pcl::LlzkToPclMode::Full, "full",
+                "Lower the IR completely."),
+               clEnumValN(::pcl::LlzkToPclMode::Stubbed, "stubbed",
+                "Replace free functions with unsupported ops with stubs.")
```
tim-hoffman:
```suggestion
                "Replace free functions containing unsupported ops with stubs.")
```

backends/pcl/lib/Conversion/PCLLoweringPass.cpp
```diff
@@ -110,6 +114,116 @@ template <typename Op> static std::string flatFullyQualifiedName(Op op) {
...
+
+  /// Maps the FQN to the operation representing it.
+  M funcs;
+
+  /// Inserts the function into the mapping.
+  void insert(FuncDefOp op) { funcs.insert({op.getFullyQualifiedName(), op}); }
+
+  /// Convenience method for the construction of the mapping.
+  void insertCallees(
+      llvm::SmallVectorImpl<FuncDefOp> &WL, mlir::SymbolTableCollection &tables, FuncDefOp f,
```
tim-hoffman: Remove `mlir::` and probably some `llvm::` prefixes since the `mlir` namespace is open

backends/pcl/lib/Conversion/PCLLoweringPass.cpp
```diff
@@ -1044,128 +1184,199 @@ class PassImpl : public pcl::impl::PCLLoweringPassBase<PassImpl> {
...
-      signalPassFailure();
-      return;
+  LogicalResult validateStructs() {
+    return failure(
+        getOperation()
+            ->walk([this](StructDefOp op) {
+      if (failed(validateStruct(op))) {
+        return WalkResult::interrupt();
+      }
+      return WalkResult::advance();
```
tim-hoffman: simplify per this constructor:
```
  /// Allow LogicalResult to interrupt the walk on failure.
  WalkResult(LogicalResult result)
      : result(failed(result) ? Interrupt : Advance) {}
```

backends/pcl/lib/Conversion/PCLLoweringPass.cpp
```diff
@@ -1044,128 +1184,199 @@ class PassImpl : public pcl::impl::PCLLoweringPassBase<PassImpl> {
...
+    }).wasInterrupted()
+    );
+  }
+
+  // PCL programs require a module-level attribute specifying the prime.
+  LogicalResult setPrime(llvm::APInt &prime) {
+    getOperation()->setAttrs(DictionaryAttr::get(&getContext()));
+    getOperation()->setAttr("pcl.prime", pcl::PrimeAttr::get(&getContext(), prime));
```
tim-hoffman, line 1315: Future: we need a string constant for "pcl.prime" since it has a lot of use locations

test/Transforms/PCLLowering/pcl_lowering_pass_stubbed.llzk
```diff
@@ -0,0 +1,208 @@
...
+// CHECK-NEXT:      %[[CONST_1:[0-9a-zA-Z_\.]+]] = pcl.const 1
+// CHECK-NEXT:      %[[CONST_65536:[0-9a-zA-Z_\.]+]] = pcl.const 65536
+// CHECK-NEXT:      %[[VAL_1:[0-9a-zA-Z_\.]+]] = pcl.lt %[[VAL_0]], %[[CONST_65536]]
+// CHECK-NEXT:      %[[VAL_2:[0-9a-zA-Z_\.]+]] = pcl.asfelt %[[VAL_1]]
+// CHECK-NEXT:      %[[VAL_3:[0-9a-zA-Z_\.]+]] = pcl.eq %[[VAL_2]], %[[CONST_1]]
+// CHECK-NEXT:      pcl.assert %[[VAL_3]]
+// CHECK-NEXT:      return 
+// CHECK-NEXT:    }
+// COM: Check that @foo is not lowered.
+// CHECK-NOT:  @foo
```
tim-hoffman: I thought it was going to leave a function without a body but this implies it's not there at all. What about calls to this function?

0xddom: That function is dead. It could be part of the witgen's call graph, but since we are removing it in this pass it's not an issue. The test below that one has a function that is actually stubbed.

backends/pcl/lib/Conversion/PCLLoweringPass.cpp
0xddom: Oh, cool!

backends/pcl/lib/Conversion/PCLLoweringPass.cpp
0xddom: Got rid of a few

test/Transforms/PCLLowering/pcl_lowering_pass_stubbed.llzk
tim-hoffman: I think I misunderstanding what you mean by a "stub" here. I assumed that meant the function has no body but all the CHECK lines have functions with bodies. I do see now that the final one has the `bit_and` removed and just returns a fresh PCL var. Can you document clearly what it means?

0xddom: I think I understand the confusion. What I'm gonna do instead is make the stub an actual stub (meaning, a function without body) and then have the translator do the pcl specific thing for such functions. Should make everything a bit more standard and flexible

tim-hoffman (approved):
LGTM

PR #686, opened 2026-08-14
Workaround for bug in upstream `remove-dead-values` pass

tim-hoffman:
See description in `lib/Dialect/POD/Transforms/PodToScalarPass.cpp`
After updating MLIR ≥22.1.0 this can be reverted because `remove-dead-values` bug is fixed and these additional passes become unnecessary (and do some reordering transformation beyond the scope of the `pod-to-scalar` pass so it would be better to remove them).


raghav198 (approved):
LGTM! Just so I understand correctly: folding the `scf.if` is just doing what the upstream dead code elimination pass _should_ be doing but currently isn't?

tim-hoffman:
> LGTM! Just so I understand correctly: folding the `scf.if` is just doing what the upstream dead code elimination pass _should_ be doing but currently isn't?

Yes, pretty much.

PR #690, opened 2026-08-18
Fix CMake when BUILD_TESTING=OFF

### LINKED ISSUE #688: [Bug]: CMake aggregate `check` target references conditionally absent test targets
### Motivation and impact

The documented aggregate `check` target cannot be invoked reliably in configurations the project supports. The build stops before any test runs, which affects developer and CI test workflows. This is a build-graph failure, not a failing test assertion.

### Bug description

At revision `683c6d412922766153a0756362762e29943cc40b`, the top-level `check` target can depend on test targets that were not created.

`check-unit` is created only when GTest is available, and `check-lit` is created only when `BUILD_TESTING` is enabled. However, the aggregate target uses:

```cmake
$<$<BOOL:BUILD_TESTING>:check-unit>
$<$<BOOL:BUILD_TESTING>:check-lit>
```

Here `BUILD_TESTING` is treated as a non-empty literal rather than the value of the CMake variable. The relevant source is [`CMakeLists.txt:394-399`](https://github.com/project-llzk/llzk-lib/blob/683c6d412922766153a0756362762e29943cc40b/CMakeLists.txt#L394-L399).

### Steps to reproduce

This minimal fixture matches the target-creation logic:

```bash
tmp=$(mktemp -d)
mkdir "$tmp/src"

cat >"$tmp/src/CMakeLists.txt" <<'EOF'
cmake_minimum_required(VERSION 3.23)
project(check_target NONE)

option(BUILD_TESTING "Build tests" ON)
set(GTest_FOUND OFF CACHE BOOL "Whether GTest is available")

if(BUILD_TESTING)
  add_custom_target(check-lit)

  if(GTest_FOUND)
    add_custom_target(check-unit)
  endif()
endif()

add_custom_target(check
  DEPENDS
    $<$<BOOL:BUILD_TESTING>:check-unit>
    $<$<BOOL:BUILD_TESTING>:check-lit>)
EOF

cmake -S "$tmp/src" -B "$tmp/off" -G Ninja \
  -DBUILD_TESTING=OFF
cmake --build "$tmp/off" --target check
```

The build fails with:

```text
ninja: error: 'check-unit', needed by 'CMakeFiles/check', missing and no known rule to make it
```

The same failure occurs with testing enabled but GTest unavailable:

```bash
cmake -S "$tmp/src" -B "$tmp/no-gtest" -G Ninja \
  -DBUILD_TESTING=ON -DGTest_FOUND=OFF
cmake --build "$tmp/no-gtest" --target check
```

The repository-faithful reproduction was confirmed in [this CI run](https://github.com/1sgtpepper/llzk-lib/actions/runs/31890102658).

### Expected behavior

The `check` target should not contain dependencies for targets that were not created.

With `BUILD_TESTING=OFF`, the project should either omit `check` or make it a clean no-op. With testing enabled but GTest unavailable, `check` should depend only on the test targets that exist, such as `check-lit`, rather than failing because `check-unit` is missing.

### LLZK version or commit

`683c6d412922766153a0756362762e29943cc40b`

### Workarounds or alternatives considered

Users can avoid the aggregate target and invoke available test targets individually, or configure with the full GTest test environment. Neither workaround makes the documented aggregate command valid in the supported configurations above.

### Additional context

The repository documents `BUILD_TESTING=OFF` as a supported configuration and documents `cmake --build . --target check` as the aggregate test command. The source also comments that test targets may not exist when testing is disabled.

- [x] I searched the open and closed issues and did not find an existing report for this problem.
- [x] I understand that external contributors should wait for the `approved` label before beginning implementation.

0xddom:
The CMake project was unbuildable if `-DBUILD_TESTING=OFF` or, if it was `ON`, if GTest was not found. This is due to the `check` catch-all target referencing targets that do not actually exists if either of the conditions above are true.

Fixes #688.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -392,11 +392,15 @@ else()
 endif()
 
 # Catch-all target for running all tests
-add_custom_target(
-  check
-  DEPENDS # test targets may not exist if BUILD_TESTING is off
-          $<$<BOOL:BUILD_TESTING>:check-unit>
-          $<$<BOOL:BUILD_TESTING>:check-lit>)
+if(BUILD_TESTING)
+  add_custom_target(
+    check
+    DEPENDS # test targets may not exist if BUILD_TESTING is off
+          $<$<BOOL:GTest_FOUND>:check-unit>
+          check-lit>)
+else()
+  add_custom_target(check) # No-op if BUILD_TESTING is disabled.
+endif()
 
 # Install
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -397,7 +397,7 @@ if(BUILD_TESTING)
     check
     DEPENDS # test targets may not exist if BUILD_TESTING is off
           $<$<BOOL:GTest_FOUND>:check-unit>
-          check-lit>)
+          check-lit)
 else()
   add_custom_target(check) # No-op if BUILD_TESTING is disabled.
 endif()
--- a/changelogs/unreleased/dani__fix-cmake-with-disabled-testing.yaml
+++ b/changelogs/unreleased/dani__fix-cmake-with-disabled-testing.yaml
@@ -1,0 +1,2 @@
+fixed:
+  - CMake misconfiguration that made the project unbuildable when BUILD_TESTING is OFF or BUILD_TESTING is ON but GTest is not found by CMake.


shankarapailoor (approved):
LGTM

PR #691, opened 2026-08-19
Support lowering `arith.select` in `--llzk-to-pcl`

### LINKED ISSUE #665: `--llzk-to-pcl` fails to legalize `arith.select` from a data-dependent `scf.if`
### Summary
`--llzk-to-pcl` rejects a module containing a *data-dependent* `scf.if` with
`failed to legalize operation 'arith.select'`. The `arith.select` is produced by canonicalization
of the `scf.if` (a ternary the PCL lowering doesn't currently support). This is the *clean-error*
counterpart to the `--llzk-full-r1cs-lowering` segfault in #664 — filing per request
so both entry points are tracked.

### Repro (minimal, 38 lines — attached below)
```
llzk-opt --inline --llzk-full-struct-inlining --llzk-to-pcl minimal.llzk
```
Output (release 3.0.0):
```
minimal.llzk:26:12: error: failed to legalize operation 'arith.select'
      %r = scf.if %b -> (!felt.type<"bn254">) {
           ^
minimal.llzk:26:12: note: see current operation: %7 = "arith.select"(%6, %4, %2)
      : (i1, !felt.type<"bn254">, !felt.type<"bn254">) -> !felt.type<"bn254">
```

### Provenance / relation to #643
The `arith.select` is **not** introduced by the `--llzk-arithmetize-scf-if` pass; `--canonicalize`
alone rewrites the input's data-dependent `scf.if` into the select (reproducible on the attached
minimal input), so it is present before any #585-stack pass runs. Invoking `--llzk-arithmetize-scf-if`
is what arithmetizes these selects away, so this is resolved by wiring #643 (+ #640) into the `to-pcl`
path; the issue documents the current gap and gives a minimal trigger.

### Expected
Either support `arith.select` in the PCL lowering, or a diagnostic directing the user to run
`--llzk-arithmetize-scf-if` first once that path is wired in.

### Environment
LLZK 3.0.0 (released) and the #643 branch built from source.

<details><summary>minimal.llzk</summary>

```mlir
// Minimal repro: a single data-dependent `scf.if` (a conditional felt select)
// inside a `constrain` body. This is the shape every EC scalar-mul ladder /
// conditional gadget lowers to. Two llzk-opt failures against release 3.0.0:
//   llzk-opt --inline --llzk-full-struct-inlining --llzk-to-pcl minimal.llzk
//       -> error: failed to legalize operation 'arith.select'
//   llzk-opt --llzk-full-r1cs-lowering minimal.llzk
//       -> SIGSEGV in PassImpl::lowerPolyToR1CS  (verified on release 3.0.0, LLVM 20.1.8)
module attributes {llzk.lang = "ACIR", llzk.main = !struct.type<@Circuit0<[]>>} {
  struct.def @Circuit0 {
    struct.member @w0 : !felt.type<"bn254"> {llzk.pub}

    function.def @compute(%arg1: !felt.type<"bn254">) -> !struct.type<@Circuit0<[]>>
        attributes {function.allow_non_native_field_ops, function.allow_witness} {
      %self = struct.new : <@Circuit0<[]>>
      %c1 = felt.const  1 : <"bn254">
      struct.writem %self[@w0] = %c1 : <@Circuit0<[]>>, !felt.type<"bn254">
      function.return %self : !struct.type<@Circuit0<[]>>
    }

    function.def @constrain(%arg0: !struct.type<@Circuit0<[]>>, %arg1: !felt.type<"bn254">)
        attributes {function.allow_constraint, function.allow_non_native_field_ops} {
      %c0 = felt.const  0 : <"bn254">
      %c1 = felt.const  1 : <"bn254">
      %b = bool.cmp eq(%arg1, %c1) : !felt.type<"bn254">, !felt.type<"bn254">
      // data-dependent select: the value of %r depends on the runtime bit %b
      %r = scf.if %b -> (!felt.type<"bn254">) {
        scf.yield %c1 : !felt.type<"bn254">
      } else {
        scf.yield %c0 : !felt.type<"bn254">
      }
      %w = struct.readm %arg0[@w0] : <@Circuit0<[]>>, !felt.type<"bn254">
      %d = felt.sub %w, %r : !felt.type<"bn254">, !felt.type<"bn254">
      constrain.eq %d, %c0 : !felt.type<"bn254">, !felt.type<"bn254">
      function.return
    }
  }
}
```
</details>

0xddom:
Add support for lowering `arith.select` ops. Fixes #665.

This PR also reorganizes the codebase for the pass to make it easier to navigate.


backends/pcl/include/pcl/Conversion/ConversionPasses.td
```diff
@@ -10,15 +10,34 @@
 #ifndef PCL_CONV_CONVERSION_PASSES_TD
 #define PCL_CONV_CONVERSION_PASSES_TD
 
-include "llzk/Pass/PassBase.td"
+include "mlir/Pass/PassBase.td"
 
-def PCLLoweringPass : LLZKPass<"llzk-to-pcl"> {
+def PCLLoweringPass : Pass<"llzk-to-pcl", "::mlir::ModuleOp"> {
```
tim-hoffman: You shouldn't need to make this change to override the `dependentDialects` field.

backends/pcl/lib/Conversion/CMakeLists.txt
```diff
@@ -3,7 +3,14 @@ add_library(LLZK::PCLConversion ALIAS PCLConversion)
 
 file(GLOB PCLConversion_SOURCES "*.cpp")
```
tim-hoffman: This glob is unused now right?

backends/pcl/lib/Conversion/PCLLoweringPass/Patterns.cpp
```diff
@@ -0,0 +1,846 @@
...
+      CallOp op, OpAdaptor adaptor, ConversionPatternRewriter &rewriter
+  ) const override {
+    SymbolTableCollection tables;
+    auto callee = op.getCalleeTarget(tables);
+    // We only care about constrain functions.
+    if (failed(callee) || !callee->get().isStructConstrain()) {
+      return failure();
+    }
+
+    auto subcmp = llvm::dyn_cast_if_present<TypedValue<StructType>>(op.getArgOperands().front());
```
tim-hoffman, line 348: This should use the adaptor to account for operands changed by other patterns.
```suggestion
    auto subcmp = llvm::dyn_cast_if_present<TypedValue<StructType>>(adaptor.getArgOperands().front());
```

backends/pcl/lib/Conversion/PCLLoweringPass/Patterns.cpp
```diff
@@ -0,0 +1,846 @@
...
+    if (!subcmp) {
+      return op->emitOpError() << "expected argument #0 to be a struct type";
```
tim-hoffman: you can simplify this a bit
```suggestion
      return op->emitOpError("expected argument #0 to be a struct type");
```

backends/pcl/lib/Conversion/PCLLoweringPass/Patterns.cpp
```diff
@@ -0,0 +1,846 @@
...
+    auto feltType = pcl::FeltType::get(rewriter.getContext());
+    SmallVector<Type> inputs(op.getNumArguments(), feltType);
+    SmallVector<Type> outputs(op.getFunctionType().getNumResults(), feltType);
+
+    auto funcOp = func::FuncOp::create(
+        op.getLoc(), flatFullyQualifiedName(op), rewriter.getFunctionType(inputs, outputs),
+        {NamedAttribute("sym_visibility", rewriter.getStringAttr("private"))}
+    );
+
+    rewriter.eraseOp(op);
```
tim-hoffman, line 488: Should be able to use `rewriter.replaceOpWithNewOp<func::FuncOp>(op, ...)` instead.

backends/pcl/lib/Conversion/PCLLoweringPass/Patterns.cpp
```diff
@@ -0,0 +1,846 @@
...
+//===----------------------------------------------------------------------===//
+
+/// Converts `struct.readm` ops that read members of felt type from the struct into `pcl.var` ops.
+struct ConvertSelfMemberReadOpOfFelt : public OpConversionPattern<MemberReadOp> {
+  using OpConversionPattern<MemberReadOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(MemberReadOp op, OpAdaptor, ConversionPatternRewriter &rewriter) const override {
+    auto parent = op->getParentOfType<FuncDefOp>();
+    if (!parent || op.getComponent() != parent.getArgument(0)) {
```
tim-hoffman, line 570: Should use `OpAdaptor` for `getComponent()` to account for possible changed value.

backends/pcl/lib/Conversion/PCLLoweringPass/Patterns.cpp
```diff
@@ -0,0 +1,846 @@
...
+      return failure();
+    }
+    if (!llvm::isa<FeltType>(defOp->get().getType())) {
+      return failure();
+    }
+
+    auto pclVar = rewriter.create<pcl::VarOp>(
+        defOp->get().getLoc(), defOp->get().getName(), defOp->get().hasPublicAttr()
+    );
+    rewriter.replaceOp(op, pclVar);
```
tim-hoffman: May be able to use `rewriter.replaceOpWithNewOp<pcl::VarOp>(op, ...)` instead.

backends/pcl/lib/Conversion/PCLLoweringPass/Patterns.cpp
```diff
@@ -0,0 +1,846 @@
...
+//===----------------------------------------------------------------------===//
+
+/// Removes `struct.readm` ops that read subcmp members from the struct.
+struct ConvertSelfMemberReadOpOfSubcmp : public OpConversionPattern<MemberReadOp> {
+  using OpConversionPattern<MemberReadOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(MemberReadOp op, OpAdaptor, ConversionPatternRewriter &rewriter) const override {
+    auto parent = op->getParentOfType<FuncDefOp>();
+    if (!parent || op.getComponent() != parent.getArgument(0)) {
```
tim-hoffman, line 600: Should use `OpAdaptor` for `getComponent()` to account for possible changed value.

backends/pcl/lib/Conversion/PCLLoweringPass/Patterns.cpp
```diff
@@ -0,0 +1,846 @@
...
+//===----------------------------------------------------------------------===//
+// ConvertSubcmpMemberReadOp
+//===----------------------------------------------------------------------===//
+
+struct ConvertSubcmpMemberReadOp : public OpConversionPattern<MemberReadOp> {
+  using OpConversionPattern<MemberReadOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(MemberReadOp op, OpAdaptor, ConversionPatternRewriter &rewriter) const override {
+    auto subcmp = llvm::dyn_cast_if_present<MemberReadOp>(op.getComponent().getDefiningOp());
```
tim-hoffman, line 654: Should use `OpAdaptor` for `getComponent()` to account for possible changed value.

backends/pcl/lib/Conversion/PCLLoweringPass/Patterns.cpp
```diff
@@ -0,0 +1,846 @@
...
+    auto defOp = op.getMemberDefOp(tables);
+    if (failed(defOp)) {
+      return failure();
+    }
+
+    auto name = (Twine(subcmp.getMemberName()) + "." + defOp->get().getName()).str();
+    auto pclVar = rewriter.create<pcl::VarOp>(
+        defOp->get().getLoc(), rewriter.getStringAttr(name), /*public=*/false
+    );
+    rewriter.replaceOp(op, pclVar);
```
tim-hoffman: May be able to use `rewriter.replaceOpWithNewOp<pcl::VarOp>(op, ...)` instead.

backends/pcl/lib/Conversion/PCLLoweringPass/Temporary.cpp
```diff
@@ -0,0 +1,93 @@
...
+    return builder.getStringAttr(name);
+  }
+
+  /// If the op is one that defines a name in the environment,
+  /// add that name to the set of used names to avoid collisions.
+  void fillUsedNames(Operation *op) {
+    TypeSwitch<Operation *>(op)
+        .Case<component::MemberDefOp>([this](auto member) {
+      usedNames.insert(member.getSymName());
+    }).Default([](auto) {});
```
tim-hoffman, line 73: `TypeSwitch` is overkill for a single case
```suggestion
    if (auto member = dyn_cast<component::MemberDefOp>(op)) {
      usedNames.insert(member.getSymName());
    }
```

backends/pcl/lib/Conversion/PCLLoweringPass/Temporary.h
```diff
@@ -0,0 +1,40 @@
...
+  llvm::DenseMap<mlir::Operation *, mlir::StringAttr> names;
+
+public:
+  Temporaries(mlir::ModuleOp root);
+
+  /// Returns, if available, the temporary created for the given operation.
+  mlir::FailureOr<mlir::StringAttr> get(mlir::Operation *op) const {
+    auto it = names.find(op);
+    if (it == names.end()) {
+      return op->emitOpError() << "does not emit a temporary";
```
tim-hoffman: could simplify
```suggestion
      return op->emitOpError("does not emit a temporary");
```

tim-hoffman (changes requested):
Why are the headers located in `lib/Conversion` rather than `include/pcl/Conversion`?

shankarapailoor:
@codex review

0xddom:
> Why are the headers located in `lib/Conversion` rather than `include/pcl/Conversion`?

Because they are private to the conversion pass. This is standard procedure.

backends/pcl/lib/Conversion/PCLLoweringPass/Temporary.cpp
0xddom: TypeSwitch compiles down to something like that so there's isn't really a performance penalty.

backends/pcl/lib/Conversion/CMakeLists.txt
0xddom: Yes.

backends/pcl/lib/Conversion/PCLLoweringPass/Patterns.cpp
0xddom: Using the adaptor breaks the tests on all these cases. I think it is because since those methods do a bunch of work behind the scenes that rely on the IR's structure, using the fake IR created with the adaptor may break some assumptions.

backends/pcl/lib/Conversion/PCLLoweringPass/Patterns.cpp
0xddom: For some reason `func::FuncOp` does not like being constructed that way. It doesn't build if I try to do it like that.

tim-hoffman (approved):
LGTM

PR #697, opened 2026-08-24
Fix in-place test-check generation

### LINKED ISSUE #689: [Bug]: generate-test-checks.py in-place mode can modify the source after generation fails
### Motivation and impact

The in-place test-check generator can damage a contributor's source file when generation fails. Repeated successful runs can also accumulate generated metadata. This affects local test files and can destroy uncommitted edits; it is not only a formatting issue.

### Bug description

At revision `683c6d412922766153a0756362762e29943cc40b`, `scripts/generate-test-checks.py --source ... -i` opens and truncates the destination source file before generation has completed.

If generation then fails, the command returns nonzero but leaves the source changed. Repeated successful in-place runs also fail to remove the previous autogenerated note, so the note is duplicated.

The relevant in-place write path is [`scripts/generate-test-checks.py:302-326`](https://github.com/project-llzk/llzk-lib/blob/683c6d412922766153a0756362762e29943cc40b/scripts/generate-test-checks.py#L302-L326).

### Steps to reproduce

Run the actual repository script:

```bash
script=scripts/generate-test-checks.py
tmp=$(mktemp -d)

cat >"$tmp/source.mlir" <<'EOF'
module {
  func.func @main() {
    %0 = "test.op"() : () -> i32
    %1 = "test.op"() : () -> i32
  }
}
EOF

cp "$tmp/source.mlir" "$tmp/source.before.mlir"
cp "$tmp/source.mlir" "$tmp/input.mlir"

set +e
python3 "$script" \
  --source "$tmp/source.mlir" \
  --source_delim_regex='^module ' \
  --variable_names='X,X' \
  -i "$tmp/input.mlir" \
  >"$tmp/stdout.log" 2>"$tmp/stderr.log"
status=$?
set -e

test "$status" -ne 0
grep -Fq "duplicate variable name" "$tmp/stderr.log"

# This comparison fails with the current implementation.
cmp "$tmp/source.before.mlir" "$tmp/source.mlir"
```

The command reports the controlled generation error, but the final `cmp` fails because the source was already truncated or rewritten before the error was raised.

The in-place update is also not idempotent:

```bash
cat >"$tmp/repeat-source.mlir" <<'EOF'
module {
  func.func @main() {
  }
}
EOF

printf '%s\n' 'module {' '}' >"$tmp/repeat-input.mlir"

python3 "$script" \
  --source "$tmp/repeat-source.mlir" \
  --source_delim_regex='^module ' \
  -i "$tmp/repeat-input.mlir"

python3 "$script" \
  --source "$tmp/repeat-source.mlir" \
  --source_delim_regex='^module ' \
  -i "$tmp/repeat-input.mlir"

test "$(grep -cF 'NOTE: Assertions have been autogenerated' \
  "$tmp/repeat-source.mlir")" -eq 1
```

The final assertion fails because the second run leaves two autogenerated note markers.

The repository-faithful reproduction was confirmed in [this CI run](https://github.com/1sgtpepper/llzk-lib/actions/runs/31890102633).

### Expected behavior

An in-place generation failure should leave the original source file unchanged.

Repeated successful in-place runs should be idempotent: previously generated assertions and the autogenerated note should be replaced rather than accumulated.

### LLZK version or commit

`683c6d412922766153a0756362762e29943cc40b`

### Workarounds or alternatives considered

Running without `-i` and reviewing the generated output before replacing the source avoids the destructive failure mode. Restoring the file from version control can recover committed files, but does not protect uncommitted edits.

### Additional context

The source update path is explicitly documented as an in-place operation. The issue is consolidated because source truncation after failure and autogenerated-note duplication are both caused by the same in-place update logic.

- [x] I searched the open and closed issues and did not find an existing report for this problem.
- [x] I understand that external contributors should wait for the `approved` label before beginning implementation.

1sgtpepper:
## Summary

Make in-place FileCheck generation safe and repeatable: replace the source only after successful generation and preserve unrelated source text on repeated runs.

## Related issues

Fixes #689.

## Changes

- Refuse dirty Git worktrees, including hidden untracked files, unless `--allow-dirty` is passed. Check the supplied and resolved source paths independently of inherited repository-local Git state.
- Use a temporary file beside the resolved source for in-place replacement, preserving permission bits and source symlinks. Reject combining in-place mode with `--output`.
- Replace generated notes and selected-prefix directives, add inline unit tests, and document the final LLZK check-layout adjustments.

## Testing

- [Fork generator and documentation checks](https://github.com/1sgtpepper/llzk-lib/actions/runs/34074052771) passed at `95d8b87882461ee0fb8ad8d0948f12cd4591a20d`: `python3 -m unittest scripts/generate-test-checks.py -v` (21 tests) and `nix --option sandbox false --print-build-logs build .#docs`.

## Submission checklist

- [x] If I am an external contributor, this PR has a linked issue marked `approved`; otherwise, this does not apply.
- [x] I added or updated tests for all relevant behavior, or explained above why tests are not needed.
- [x] I updated the relevant TableGen or other documentation, or explained above why documentation is not needed.
- [x] I added a changelog entry describing user-visible changes. (`create-changelog` in the Nix development shell creates a template.)
- [x] I enabled **Allow edits from maintainers** if this PR comes from a fork.

## AI assistance

- [ ] No AI tools contributed to this PR.
- [x] AI tools contributed to this PR.

**Tools used:** AI coding assistant

**How the tools contributed:** Implementation and code review.

**How I verified the contribution:** Reviewed the final diff and exact-head hosted results.


scripts/generate-test-checks.py
```diff
@@ -281,7 +348,18 @@ def main():
...
+        "--inplace",
+        action="store_true",
+        default=False,
+        help="Update --source in place after successful generation.",
+    )
+    parser.add_argument(
+        "--allow-dirty",
+        action="store_true",
+        help="Allow in-place updates in a dirty Git worktree.",
+    )
```
0xddom, line 443: This argument should default to `False`

scripts/generate-test-checks.py
```diff
@@ -422,8 +534,37 @@ def main():
...
+                    os.unlink(temporary_source)
+                except FileNotFoundError:
+                    pass
+    else:
+        output.close()
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main())
```
0xddom, line 698: If we are going to directly exit using main's return value then main should return 0 at the end. Currently it doesn't and a function that does not have a `return` statement at the end of its body implicitly returns `None`.

scripts/generate-test-checks.py
```diff
@@ -422,8 +534,37 @@ def main():
...
+        output.close()
+        temporary_source = None
+        try:
+            with tempfile.NamedTemporaryFile(
+                mode="w",
+                dir=os.path.dirname(source_path),
+                prefix="." + os.path.basename(source_path) + ".",
+                suffix=".tmp",
+                delete=False,
+            ) as temporary:
```
0xddom: Instead of buffering everything into a `StringIO`, use the temporary file directly. Then do the moving at the end.

test/Transforms/generate-test-checks-inplace.llzk
0xddom, line 1: I don't like this file for two reasons. 1) We shouldn't have a lit test for what is essentially a dev tool, and 2) this is going to be a nightmare to maintain. 

Please move these tests into the script itself as unit-tests below the `if __name__ == "__main__":` block. The idea is to have the script run its normal operation in the `main` function when invoked as a script and to run the unit tests when invoked with `python -m unittest`.

Then we can add a CI pipeline that runs the tests if-and-only-if there are changes on the script.

0xddom (changes requested):
I have no way of knowing if the tests are testing the right thing since they are an unreadable mess. Please work on that.

PR #698, opened 2026-08-24
New verif ops

0xddom:
This PR adds ops to the verif dialect with the idea of lowering them to their corresponding PCL ops. The conversion patterns for that will get added on a separate PR after #691 is merged.

### DIFF
--- a/changelogs/unreleased/dani__verif-det-extensions.yaml
+++ b/changelogs/unreleased/dani__verif-det-extensions.yaml
@@ -1,3 +1,3 @@
 added:
-  - Operations `verif.det.prove` and `verif.det.assume` for better support determinism checkers.
+  - Operations `verif.det.prove` and `verif.det.assume` to better support determinism checkers.
   - Attribute `function.allow_verif_ops` for the `function.def` op that allows inlining some `verif` dialect ops inside `function.def` ops. 
--- a/changelogs/unreleased/dani__verif-det-extensions.yaml
+++ b/changelogs/unreleased/dani__verif-det-extensions.yaml
@@ -1,0 +1,3 @@
+added:
+  - Operations `verif.det.prove` and `verif.det.assume` for better support determinism checkers.
+  - Attribute `function.allow_verif_ops` for the `function.def` op that allows inlining some `verif` dialect ops inside `function.def` ops. 
--- a/include/llzk/Dialect/Function/IR/OpTraits.h
+++ b/include/llzk/Dialect/Function/IR/OpTraits.h
@@ -24,7 +24,7 @@ mlir::LogicalResult verifyNotFieldNativeTraitImpl(mlir::Operation *op);
 /// is inside a `function.def` operation.
 ///
 /// For example, `Verification<WitnessGen>` will call this function passing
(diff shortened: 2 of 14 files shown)


changelogs/unreleased/dani__verif-det-extensions.yaml
```diff
@@ -0,0 +1,3 @@
+added:
+  - Operations `verif.det.prove` and `verif.det.assume` for better support determinism checkers.
```
tim-hoffman:
```suggestion
  - Operations `verif.det.prove` and `verif.det.assume` to better support determinism checkers.
```

include/llzk/Dialect/Verif/IR/Ops.td
```diff
@@ -86,26 +142,39 @@ class RequireOpBase<string mnemonic, list<Trait> traits = []>
...
 
-class EnsureOpBase<string mnemonic, list<Trait> traits = []>
-    : ContractConditionOp<mnemonic, traits#[PostconditionOpInterface]>;
+class EnsureOpBase<string mnemonic, bit inlinable = false,
+                   list<Trait> traits = []>
+    : ContractConditionOp<"ensure_"#mnemonic, inlinable,
+                          traits#[PostconditionOpInterface]> {
+  let description = [{
+    Encodes a postcondition in the `@}]#mnemonic#[{` function.
+  }]#!if(inlinable, [{
```
tim-hoffman, line 167: It seems like the use of `inlinable` here just happens to work but it would be more accurate to check if `traits` contains a `Verification` instance. Or perhaps you refactor `EnsureOpBase` to have an additional `list<string>` whose emptiness is used as the condition here and, if non-empty, is used to add the necessary `Verification` trait.

include/llzk/Dialect/Function/IR/OpTraits.h
```diff
@@ -18,6 +18,8 @@ namespace llzk::function {
 mlir::LogicalResult verifyConstraintGenTraitImpl(mlir::Operation *op);
 mlir::LogicalResult verifyWitnessGenTraitImpl(mlir::Operation *op);
 mlir::LogicalResult verifyNotFieldNativeTraitImpl(mlir::Operation *op);
+mlir::LogicalResult
+verifyVerificationTraitImpl(mlir::Operation *op, llvm::function_ref<mlir::LogicalResult()>);
```
tim-hoffman, line 30: Can you document this function, being sure to mention that it's only called when the op has a `FuncDefOp` parent.

tim-hoffman (changes requested):
Could you also add lit tests for "det.prove" and "det.assume"

include/llzk/Dialect/Verif/IR/Ops.td
0xddom: I went with both the boolean and the list of strings, to be more explicit. And the way I understand it, having extra traits tied to `Verification` is not a requirement for having that trait. But it just so happens that the one where I made those changes needs `ConstraintGen`. If you look at the `verif.det.*`, for example, they have `Verification<>` because they don't need extras.

tim-hoffman (approved):
LGTM

PR #702, opened 2026-08-26
Ensure `tvar` symbol refers to template param only

tim-hoffman:


lib/Util/SymbolHelper.cpp
```diff
@@ -493,7 +493,19 @@ LogicalResult verifyTypeResolution(SymbolTableCollection &tables, Operation *ori
...
-    return verifyParamOfType(tables, vTy.getNameRef(), vTy, origin);
+    // Unlike other type parameters, a type variable may only name a parameter
+    // of the enclosing template; it cannot resolve to a global.
+    FailureOr<TemplateOp> parent = getConstResolutionTemplate(tables, origin);
+    if (failed(parent)) {
+      return failure();
+    }
+    TemplateOp templateOp = *parent;
+    if (templateOp && templateOp.getConstNamed<TemplateParamOp>(vTy.getNameRef())) {
+      return success();
```
0xddom, line 504: Should we still do the function call that was removed but on line 504?

tim-hoffman: No, this new implementation is essentially the first part of `verifyParamOfType` without the remaining cases that would normally be allowed but are not valid for template parameters.

0xddom (approved):

PR #705, opened 2026-08-28
Convert new verif ops to PCL

0xddom:
Adds support in `--llzk-to-pcl` for converting the latest `verif` dialect ops to PCL.


tim-hoffman (approved):
LGTM

PR #710, opened 2026-09-01
Make LLZK->PCL pass fail on structs with no outputs

0xddom:


changelogs/unreleased/dani__llzk-to-pcl-fail-on-no-output.yaml
```diff
@@ -0,0 +1,2 @@
+changed:
+  - "`--llzk-to-pcl` pass now fails if any of the structs in the circuit has not public members"
```
tim-hoffman:
```suggestion
  - "`--llzk-to-pcl` pass now fails if any of the structs in the circuit have no public members"
```

tim-hoffman (changes requested):
LGTM; changelog wording just needs the clarification suggested

changelogs/unreleased/dani__llzk-to-pcl-fail-on-no-output.yaml
0xddom: Good catch...

tim-hoffman (approved):

PR #712, opened 2026-09-01
Disallow wildcard as struct parameter

tim-hoffman:
This restriction should have been added originally when wildcard support was added. There is no reason to use the `?` in that position.

### DIFF
--- a/unittests/Util/TypeHelperTest.cpp
+++ b/unittests/Util/TypeHelperTest.cpp
@@ -47,13 +47,14 @@ TEST_F(TypeHelperTests, test_arrayTypesUnify_withDynamic_2) {
 }
 
 TEST_F(TypeHelperTests, test_structTypesUnify) {
+  // StructType itself cannot be created with `?` in its parameter list.
+  // Instead test a nested ArrayType as TypeAttr in the list passes.
   IndexType tyIndex = IndexType::get(&ctx);
-  Attribute i1 = IntegerAttr::get(tyIndex, 128);
-  Attribute i2 = IntegerAttr::get(tyIndex, ShapedType::kDynamic);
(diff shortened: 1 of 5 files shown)


unittests/Util/TypeHelperTest.cpp
```diff
@@ -47,13 +47,14 @@ TEST_F(TypeHelperTests, test_arrayTypesUnify_withDynamic_2) {
...
-  Attribute i2 = IntegerAttr::get(tyIndex, ShapedType::kDynamic);
-  StructType a = StructType::get(FlatSymbolRefAttr::get(&ctx, "TheName"), ArrayRef {i1});
-  StructType b = StructType::get(FlatSymbolRefAttr::get(&ctx, "TheName"), ArrayRef {i2});
-  // `false` because StructType does not allow `kDynamic`
-  ASSERT_FALSE(structTypesUnify(a, b));
+  Attribute t1 = TypeAttr::get(ArrayType::get(tyIndex, {242, ShapedType::kDynamic}));
+  Attribute t2 = TypeAttr::get(ArrayType::get(tyIndex, {ShapedType::kDynamic, 5}));
+  StructType a = StructType::get(FlatSymbolRefAttr::get(&ctx, "TheName"), ArrayRef {t1});
+  StructType b = StructType::get(FlatSymbolRefAttr::get(&ctx, "TheName"), ArrayRef {t2});
+  ASSERT_TRUE(structTypesUnify(a, b));
```
0xddom, line 57: So, `?` is not allowed as a parameter, but it is allowed inside other parametrizable types used as parameter, correct?

tim-hoffman: Correct. This PR rejects `struct.type<[?]>` but not `struct.type<[array.type<? x felt>]>` which does also seem like it should be rejected. I'll give it more thought and make a separate PR if needed.

0xddom (approved):

PR #717, opened 2026-09-03
Normalize `global.def` types

tim-hoffman:
Update `global.def` parser to make the optional initializer attribute have the same type as the def itself and the verifier to enforce that. This avoids the need for a lot of special cases and checks in passes, etc. that deal with these because there is one source of truth on the type. The parser has an additional special case for updating the `global.def` type when it's an unspecified-field `felt.type` and the initializer attribute specifies a field.


lib/Dialect/Global/IR/Dialect.cpp
```diff
@@ -9,22 +9,63 @@
...
+          return InFlightDiagnosticWrapper::createSilent(context);
+        }
+        );
+        if (failed(normalized)) {
+          return global.emitError(
+              "contains a legacy initializer that is incompatible with its declared type"
+          );
+        }
+        global.setInitialValueAttr(*normalized);
+        global.setTypeAttr(TypeAttr::get(type));
```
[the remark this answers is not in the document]
tim-hoffman: Very low likelihood this will be encountered. Will not fix now.

lib/Dialect/Global/IR/Ops.cpp
```diff
@@ -30,49 +32,204 @@ using namespace llzk::string;
...
+            normalizeGlobalInitializer(elementType, element, emitError);
+        if (failed(normalized)) {
+          return failure();
+        }
+        elements.push_back(*normalized);
+      }
+      return ArrayAttr::get(value.getContext(), elements);
+    }
+  }
+  return value;
```
0xddom: This function is a bit messy IMO. I think it could benefit from using `llvm::TypeSwitch` or from splitting it into smaller pieces.

Also, some cases like `StringType`'s look redundant, as in the returned attribute would effectively be equal to `value`.

lib/Dialect/Global/IR/Ops.cpp
```diff
@@ -30,49 +32,204 @@ using namespace llzk::string;
...
+  return success();
+}
 
-  // Special case for parsing LLZK FeltType to match format of FeltConstantOp.
-  // Not actually necessary but the default format is verbose. ex: "#felt<const 35>"
-  if (isa<FeltType>(specifiedType)) {
-    FeltConstAttr feltConstAttr;
-    if (parser.parseCustomAttributeWithFallback<FeltConstAttr>(feltConstAttr)) {
+/// Parse an initializer attribute recursively so felt values retain their optional field syntax
+/// even when nested in an array.
```
0xddom: Wouldn't this allow for different literals inside the array having different fields?

lib/Dialect/Global/IR/Ops.cpp
```diff
@@ -30,49 +32,204 @@ using namespace llzk::string;
...
+
+  Attribute initialValue;
+  if (succeeded(parser.parseOptionalEqual())) {
+    SMLoc initializerLoc = parser.getCurrentLocation();
+    if (failed(parseInitialValueForType(parser, declaredType, initialValue))) {
+      return failure();
+    }
+    if (failed(normalizeParsedInitialValue(parser, initializerLoc, declaredType, initialValue))) {
+      return failure();
+    }
```
0xddom: nit

```suggestion
    if (failed(parseInitialValueForType(parser, declaredType, initialValue)) || 
        failed(normalizeParsedInitialValue(parser, initializerLoc, declaredType, initialValue))) {
      return failure();
    }
```

lib/Dialect/Global/IR/Ops.cpp
```diff
@@ -30,49 +32,204 @@ using namespace llzk::string;
...
-    // Special case for LLZK FeltType to match format of FeltConstantOp.
-    // Not actually necessary but the default format is verbose. ex: "#felt<const 35>"
-    if (FeltConstAttr feltConstAttr = llvm::dyn_cast<FeltConstAttr>(initialValue)) {
-      p.printStrippedAttrOrType<FeltConstAttr>(feltConstAttr);
-    } else {
-      p.printAttributeWithoutType(initialValue);
-    }
+    printInitialValue(p, initialValue);
   }
+  p.printOptionalAttrDict((*this)->getAttrs(), {"constant", "sym_name", "type", "initial_value"});
```
0xddom, line 271: Why `(*this)->`?

lib/Dialect/Global/IR/Ops.cpp
```diff
@@ -145,6 +312,21 @@ LogicalResult ensureAttrTypeMatch(
...
         );
       }
+      if (auto feltElemTy = llvm::dyn_cast<FeltType>(arrTy.getElementType())) {
+        for (Attribute element : arrVal) {
+          if (auto feltValue = llvm::dyn_cast<FeltConstAttr>(element)) {
+            FeltType valueType = feltValue.getType();
+            if (valueType.hasField()) {
+              if (feltElemTy.hasField() && feltElemTy != valueType) {
+                return errFn().append(
+                    "initializer array contains conflicting types ", valueType, " vs ", feltElemTy
```
0xddom: Ah, ok. Even if the parser allows it the verifier will catch it.

lib/Dialect/Global/IR/Ops.cpp
```diff
@@ -145,6 +312,21 @@ LogicalResult ensureAttrTypeMatch(
...
+                );
+              }
+              feltElemTy = valueType;
+            }
```
0xddom: nit
```suggestion
            if (!valueType.hasField()) {
              continue;
            }
            if (feltElemTy.hasField() && feltElemTy != valueType) {
              return errFn().append(
                  "initializer array contains conflicting types ", valueType, " vs ", feltElemTy
              );
            }
            feltElemTy = valueType;
```

test/Dialect/Global/globals_fail.llzk
```diff
@@ -192,6 +229,14 @@ module attributes {llzk.lang} {
 
 // -----
 
+// TEST: index array initializers reject narrow negative values rather than changing their value.
+module attributes {llzk.lang} {
+  // expected-error@+1 {{negative narrow integer initializer cannot be converted to `index`}}
+  global.def const @g : !array.type<1 x index> = [-1 : i8]
```
0xddom, line 235: Can't we just sign-extend the narrower value? Or is `index` supposed to be unsigned? I can't recall from the top of my head.

0xddom (approved):
I just have some nits about deeply nested code, other than that LGTM

lib/Dialect/Global/IR/Ops.cpp
tim-hoffman: The StringAttr parse with type `none` for some reason which triggers a failure in the verification expecting StringType. It's present in the old-version bytecode this way as well so rather than updating the parser, handling in the normalize function covers both cases.

lib/Dialect/Global/IR/Ops.cpp
tim-hoffman: It can't implicitly get `Operation*` from `GlobalDefOp*` so there's an `error: no member named 'getAttrs' in 'llzk::global::GlobalDefOp'`.

0xddom (approved):

test/Dialect/Global/globals_fail.llzk
tim-hoffman: The `index` and `i*` types are signless so `i8` carries no information to distinguish `-1` from `255`. I added a test case to show that they both hit this same error case (which I reworded appropriately).

tim-hoffman: There's no reason for someone to ever write IR in this way so the error is acceptable. It should be `!array.type<1 x index> = [-1]` or `!array.type<1 x index> = [-1 : index]` instead.

lib/Dialect/Global/IR/Ops.cpp
```diff
@@ -30,49 +32,243 @@ using namespace llzk::string;
...
-  if (isa<FeltType>(specifiedType)) {
-    FeltConstAttr feltConstAttr;
-    if (parser.parseCustomAttributeWithFallback<FeltConstAttr>(feltConstAttr)) {
+/// Parse an initializer attribute recursively so felt values retain their optional field syntax
+/// even when nested in an array. If there is a conflict among the felt values, the verifier will
+/// catch it later.
+static ParseResult parseInitialValueForType(OpAsmParser &parser, Type type, Attribute &value) {
+  if (llvm::isa<FeltType>(type)) {
+    FeltConstAttr feltValue;
+    if (parser.parseCustomAttributeWithFallback<FeltConstAttr>(feltValue)) {
```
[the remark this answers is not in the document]
tim-hoffman: Not supporting this case. Backwards incompat. is acceptable.

shankarapailoor (approved):
LGTM. I additionally had codex run a review and it found no new issues. 

As an aside, do you think global initialization could be typed IR like:

```
global.def @x : !felt.type<"babybear"> {
  %value = felt.const 35 : !felt.type<"babybear">
  global.yield %value : !felt.type<"babybear">
}
```

We would then have the type system enforce this, but maybe it isn't needed?

PR #634, opened 2026-07-23
Verify felt template argument fields

1sgtpepper:
## Summary

Verify explicit and inferred template arguments for struct types, free-function calls, and contract inclusions. Accept compatible fieldless felt values in the required field and diagnose conflicting values, incompatible fields, and invalid symbol references.

## Related issues

Part of #479; split from #604. Related: #718, #719, and #721.

## Changes

- Share argument verification between `function.call` and `verif.include`, including declaration notes and checks for every inferred felt candidate.
- Preserve template and included-module context when resolving arguments, so unrelated same-spelled symbols cannot satisfy inference.
- Document accepted argument forms and distinguish fieldless restrictions from restrictions requiring an explicit field.

## Testing

Regression: Run `llzk-opt -split-input-file -verify-diagnostics test/Dialect/Function/call_with_symbolic_felt_conflict_fail.llzk` with [this fixture](https://github.com/1sgtpepper/llzk-lib/blob/e942a8399e728cc6f428a517c8abd24fd203dcb8/test/Dialect/Function/call_with_symbolic_felt_conflict_fail.llzk). Its `equivalent` split must accept a global and literal denoting the same bn128 value; conflicting values or fields must be diagnosed. At `ba936ff5`, the fixture fails its expectations and the repeated-candidate verifier crashes ([baseline output](https://github.com/1sgtpepper/llzk-lib/actions/runs/33991545416)).

- [Fork validation](https://github.com/1sgtpepper/llzk-lib/actions/runs/34045881121) passed: `nix build` for `.#debugGCC`, `.#debugClangCov`, `.#checks.x86_64-linux.llzk-installcheck-debug`, and `.#checks.x86_64-linux.llzk-installcheck-release`; `nix --option sandbox false --print-build-logs build .#docs`.
- [Fixture replay](https://github.com/1sgtpepper/llzk-lib/actions/runs/34045921690): test commands passed; output-comparison differences were reviewed manually (replay job remains failed).
- [Style checks](https://github.com/1sgtpepper/llzk-lib/actions/runs/34045879161): `git-clang-format` 20.1.8 passed; `clang-tidy` reported no new LLZK source diagnostics.

## Submission checklist

- [x] If I am an external contributor, this PR has a linked issue marked `approved`; otherwise, this does not apply.
- [x] I added or updated tests for all relevant behavior, or explained above why tests are not needed.
- [x] I updated the relevant TableGen or other documentation, or explained above why documentation is not needed.
[...]

PR #718, opened 2026-09-04
tighten restrictions on symbol template arguments

tim-hoffman:
Split separate concern from https://github.com/project-llzk/llzk-lib/pull/634: Require symbols used as template arguments to resolve to a `poly.param` or `poly.expr` binding or a constant global.


raghav198 (approved):
LGTM

PR #721, opened 2026-09-04
Add shared ODS for CallOp and IncludeOp

tim-hoffman:
`IncludeOp` was originally just a copy of `CallOp` but it's better to avoid so much duplicate code and just make a shared interface.


tim-hoffman:
@codex review

PR #723, opened 2026-09-10
Resolve paths relative to `include.from` if necessary

tim-hoffman:
Split separate concern from https://github.com/project-llzk/llzk-lib/pull/634: Resolve paths in included function and contract signatures relative to the included module when verifying explicit or inferred template arguments.


lib/Dialect/Function/IR/Ops.cpp
```diff
@@ -800,7 +800,8 @@ struct KnownTargetVerifier : public CallOpVerifier {
 
       // Check that the provided instantiation values are consistent with what type unification
       // of the target function types against the call's operand and result types would determine.
-      FailureOr<UnificationMap> unifyResult = callOp->unifyTypeSignature(tgtType);
+      FailureOr<UnificationMap> unifyResult =
+          callOp->unifyTypeSignatureWithNamespace(tgtType, includeSymNames);
       assert(succeeded(unifyResult) && "already checked by `verifyInputs()` and `verifyOutputs()`");
```
shankarapailoor: Is the assert underneath correct? I took a look at the verifier for callOp and it doesn't return early but just aggregates failures. Might be worth fixing in this PR.

shankarapailoor (commented):
Looks mostly good but just have a comment

lib/Dialect/Function/IR/Ops.cpp
tim-hoffman: This is referring to https://github.com/project-llzk/llzk-lib/blob/d1198631dbe9906cb149d1ddcd0dd1ca071dae6b/lib/Dialect/Function/IR/Ops.cpp#L734-L740
If the `callOp` arg operands match the `tgtType` inputs and the `callOp` result types match the `tgtType` results then the complete type signature must unify.

shankarapailoor: My point was that at this point in the verification it hasn't been established that `verifyInputs()` holds because the errors get aggregated right? So the assertion message seems wrong there.

tim-hoffman: I'll double check

tim-hoffman: You're right. I didn't look back at the top-level caller when I checked this yesterday.

tim-hoffman:
@codex review

shankarapailoor (approved):
LGTM

PR #727, opened 2026-09-14
Allow `global.read` with new `const` modifier within a `poly.expr` initializer

tim-hoffman:
Add optional `const` modifier to `global.read` op and verify that it targets a `global.def` with the `const` modifier when used. This allows `global.read` to be used within a `poly.expr` initializer with only op-local verifications.

This allows large array literals that are read within a `poly.expr` to be defined more succinctly as a constant `global.def`.


lib/Dialect/Global/IR/Ops.cpp
```diff
@@ -433,13 +438,59 @@ verifySymbolUsesImpl(GlobalRefOpInterface refOp, SymbolTableCollection &tables)
...
     return failure();
   }
+  if (isConstant() && !tgt->get().isConstant()) {
+    return emitOpError().append(
+        "marked as 'const' can only target '", GlobalDefOp::getOperationName(),
+        "' marked as 'const'"
+    );
+  }
+  if (!isConstant() && tgt->get().isConstant()) {
+    emitWarning()
```
raghav198: Maybe I'm missing some context, but if marking a read op as `const` has different semantics, why is this a warning and not an error? (And if it doesn't, why is it a warning at all?)

raghav198 (commented):
One minor comment, otherwise LGTM

lib/Dialect/Global/IR/Ops.cpp
tim-hoffman: Fair question. Adding `const` to `read` signifies that the target `def` is immutable. The verifier ensures that `const read` must target a `const def`.
However, if you don't put `const` on the `read`, the compiler cannot make any assumption about the `def`, it may be mutable or immutable.

tim-hoffman: Warning may be too strong. Maybe I can lower it to remark level. I basically wanted to say, "hey, you may be more optimal results if you also mark this `const`."

raghav198: Ah, as in it signals `const`-ness to the compiler even when the actual def isn't visible? Yeah maybe remark is appropriate if its just about optimization opportunities and not potentially something incorrect.

tim-hoffman: Exactly. The important part is, visible or not, the verifier does not need to lookup the symbol at all to perform the check.

test/Dialect/Poly/template_fail.llzk
```diff
@@ -121,12 +121,27 @@ module attributes {llzk.lang} {
...
 // -----
 
+// TEST: `global.read` without `const` cannot be used in a template expression
+module attributes {llzk.lang} {
+  global.def const @b : !felt.type = 999
+
+  poly.template @T4 {
+    poly.expr @D {
+      // expected-error@+1 {{'global.read' op without `const` is not allowed within a `poly.expr` initializer}}
+      %c = global.read @b : !felt.type
```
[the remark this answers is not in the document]
tim-hoffman: Invalid. Test passes as written.

lib/Dialect/Polymorphic/IR/Ops.cpp
```diff
@@ -149,21 +149,36 @@ LogicalResult TemplateExprOp::verifyRegions() {
...
+                           << TemplateExprOp::getOperationName() << "` initializer"
+      );
+    };
+    // Note: If memory-effect traits are added to call ops in the future,
+    // this check should be updated to check for those traits as well.
+    if (llvm::isa<function::CallOp>(p)) {
+      emitIllegalOpError(Twine("'") + p->getName().getStringRef() + "' op");
+      return WalkResult::interrupt();
+    }
+    if (llvm::isa<global::GlobalRefOpInterface>(p) && !isMemoryEffectFree(p)) {
```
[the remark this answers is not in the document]
tim-hoffman: Flattening handling can be done in a separate PR.

raghav198 (approved):
LGTM

PR #739, opened 2026-09-16
Clarify that read/write ops copy rather than create an alias

tim-hoffman:
- Clarified documentation for read-like and write-like ops to specify copy semantics.
 - Fix bug in `llzk-redundant-read-write-pass` where it used aliasing rather than copy semantics for aggregate values.


tim-hoffman:
@codex review

tim-hoffman:
@codex review

tim-hoffman:
@codex review

shankarapailoor (changes requested):
The fixes mostly look good but there are a couple issues with respect to `cast` ops that I found with Claude's help. 

Per my understanding, cast aliasing is modeled purely by two `ValueMap` keys sharing one `ReferenceNode`, and Claude pointed me to two places where that sharing silently breaks while both keys keep their facts.

### 1. `intersectValueMap` severs cast alias groups at region joins

`cloneValueMap` dedupes so that `%input`/`%cast` keep sharing one node across clones, but `intersectValueMap` calls `greatestCommonSubtree` per key with no memoization. Both keys map to the same `(parent, region)` tree pair at a join, yet each call allocates a fresh node — so after any region op the group is severed. A later write through one alias is then invisible to reads through the other. Here's a concrete example Claude gave me. 

```mlir
%cast = poly.unifiable_cast %input : (!array.type<1 x index>) -> !array.type<1 x index>
scf.if %cond { }                  // any region op severs the group at the join
%r1 = array.read %input[%zero] : !array.type<1 x index>, index
array.write %cast[%zero] = %v : !array.type<1 x index>, index   // updates only %cast's tree
%r2 = array.read %input[%zero] : !array.type<1 x index>, index  // RAUW'd to %r1 — wrong, actual value is %v
```
Claude's initial proposed fix was to memoize gcs results on the `(lhs.get(), rhs.get())` pair inside `intersectValueMap`, mirroring `cloneValueMap`'s dedup. However, I was thinking that since cast aliasing is a flow-invariant of the def-use graph, an alternative that removes the bookkeeping entirely is canonicalizing map keys through `getAggregateAliasRoot` which means that a group only ever occupies one slot, and the cast handler / clone dedup / unknown-effect group-erasure loop all become unnecessary.

### 2. `mayMutateThroughAlias` only walks casts forward from the queried value

I think `mayMutateThroughAlias` also doesn't handle `cast`s correctly. From what I understand, the the way the function works is it gets the uses of the queried value but that queried value may itself have aliases (like with casts). Here's the shape of the cex I have in mind, 

```mlir
...
pod.write %pod[@v] = %cast : ...      // record snapshotSource = {%cast, W1}
array.write %arr[%c0] = %b : ...      // mutates the shared aggregate — not a use of %cast, unseen
pod.write %pod[@v] = %cast : ...      // isRedundantAggregateWrite finds no mutation → erased. Wrong:
```

I think the PR will drop the second write as redundant because it misses that `%cast` and `%arr` are aliased.

tim-hoffman:
@codex review

lib/Transforms/LLZKRedundantReadAndWriteEliminationPass.cpp
```diff
@@ -737,14 +935,25 @@ class PassImpl : public llzk::impl::RedundantReadAndWriteEliminationPassBase<Pas
       return v;
     };
 
+    // Transparent aggregate casts preserve identity. Store one tree under the
+    // root value for the whole cast chain, rather than one entry per SSA alias.
+    auto canonicalStateKey = [&](Value v) { return getAggregateAliasRoot(translate(v)); };
```
shankarapailoor: Claude found a missed optimization opportunity. Looking into it, it seems that translate is applied only to `v`, but `getAggregateAliasRoot`
then walks the raw cast operands without translating them. So for `%c = cast %r` where `%r`
is replacement-pending (`translate(%r) = %prior`), uses of `%c` key into `%r`'s slot while
the group's facts live under `%prior`'s slot which is a missed fact, e.g. a read through `%c` won't
reuse facts learned through `%prior`.

shankarapailoor (commented):
It looks mostly good to me, but Claude found one optimization opportunity you should look at.

shankarapailoor (approved):
LGTM

